Issue #94, opened 2022-10-10
[Important] Simulation inconsistency in GUI due to thread racing condition


cncastillo:
If you run the simulation multiple times in the GUI there are some inconsistencies in the phase. Plotting the real value of the signal below:

![image](https://user-images.githubusercontent.com/5957134/194873159-14e5362e-74cb-4daa-86ed-96bb71022bed.png)

This also affects the reconstructed phase:

![image](https://user-images.githubusercontent.com/5957134/194873327-4a87a3de-cfcc-4697-8a6b-b4967f79150a.png)

Possible causes:
 - 1. The raw data is modified each run. **Solution**: change those lines to use "copies" of the object instead.
 - 2. The interpolation object is being modified for each thread, causing problems. **Solution**: remove interpolation.
 - 3.  Synchronization problems when dealing with threads? **Solution**: Use @sync macro (?)

We must check if this also happens when simulating normally (no UI, case 2 and 3).

cncastillo:
I will try to remove the interpolation because I think is causing more problems.

cncastillo:
After debugging a little bit, I realized that the problem was not in the boundary of the simulation blocks, but within the blocks:
![image](https://user-images.githubusercontent.com/5957134/195070555-511222d2-6ac5-4f22-b147-9e4130edf550.png)
Meaning that the problem was probably caused by iii. The operation `S .+= aux` in `run_spin_precession_parallel` was producing the wrong answer due to thread [racing conditions](https://docs.julialang.org/en/v1/manual/multi-threading/#Atomic-Operations).

One solution is to do something like
![image](https://user-images.githubusercontent.com/5957134/195071226-dca39b6f-70a7-4d30-8e44-2b9c389d776d.png)
where we store the results per thread. I have verified that this solves the problem, but also uses more memory and I think is slower.

cncastillo:
While testing I realized this solution is only valid if we use `Nthreads`, for simParams with `Nthreads=1` it fails, as the `threadid` does not correspond with the dimension of `S`.

cncastillo:
I think this is the best solution so far (using `ThreadsX`, [more info](https://www.youtube.com/watch?v=ZYzlhp-W0a8))

```julia
S = ThreadsX.mapreduce(+, parts) do p #Thread-safe summation
    S_p, M0[p] = run_spin_precession(obj[p],seq,t,Δt; M0=M0[p], gpu)
    S_p
end
```

PR #96, opened 2022-10-13
Removed simulation inconsistency and other fixes

cncastillo:
This branch was used to fix issue #94, and was called remove_interpolation, as I thought that was the problem. In the end, it was a problem with the thread safety of some operations. I also fixed some bugs and added some improvements to the plotting functions, like the ability to plot the simulation blocks and sequence blocks in `plot_seq` and `plot_signal`.


cncastillo:
Oops, deleted Supressor and it was being used in the Tests. This will be fixed shortly.

cncastillo:
The solution was to create a test environment ([more info](https://pkgdocs.julialang.org/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above)). Also, now we are using `ThreadsX` to do fast and thread-safe operations.

PR #104, opened 2022-10-18
Fix bug for Reconstruct button

beorostica:
If a user press the button Reconstruct! without previously pressing the Simulate! button, then the julia REPL throw an error. This change overcomes this problem.


cncastillo:
I do not like this solution. You should just change the default raw_ismrmd in KomaUi.jl.

PR #111, opened 2022-10-26
Add 3D simulation example

beorostica:
Add 3D simulation example


beorostica:
Highlight that a voxel represents a spin density!!!

PR #169, opened 2023-05-26
Added workaround to use KomaMRI in Julia 1.9

beorostica:
I added an important note in the readme and in the docs to use KomaMRI in Julia 1.9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,12 @@ To install just write the following in the Julia REPL:
 ```julia
 ] add KomaMRI
 ```
+
+**IMPORTANT!!!** After Julia 1.9 release, some features don't work properly when using multithreading with the user interface. By the time being, the following workaround is necessary to solve some errors:
+```julia
+] add WebIO#a18e830a5a4d1193472e2a111ca4a62c1165f716
+```
+
 ## First run
 KomaMRI.jl comes with a handy GUI that contains a brain phantom with an EPI sequence. To open it use:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -85,6 +85,12 @@ To install just write the following in the Julia REPL:
 ```julia
 ] add KomaMRI
 ```
+
+**IMPORTANT!!!** After Julia 1.9 release, some features don't work properly when using multithreading with the user interface. By the time being, the following workaround is necessary to solve some errors:
+```julia
+] add WebIO#a18e830a5a4d1193472e2a111ca4a62c1165f716
+```
```
cncastillo, line 92: **IMPORTANT!!!** In Julia 1.9, some UI features don't work properly when using Julia with multiple threads (by starting it with `julia -t auto` or `julia -t Nthreads`). This is currently being fixed. For the time being, we recommend installing a dev version of `WebIO.jl`  that fixes these problems. This can be done by doing the following:
```julia
] add WebIO#a18e830a5a4d1193472e2a111ca4a62c1165f716
```

docs/src/getting-started.md
```diff
@@ -13,6 +13,12 @@ julia> ]
 
 (@v1.8) pkg> add KomaMRI
 ```
+
+**IMPORTANT!!!** After Julia 1.9 release, some features don't work properly when using multithreading with the user interface. By the time being, the following workaround is necessary to solve some errors:
+```julia
+(@v1.9) pkg> add WebIO#a18e830a5a4d1193472e2a111ca4a62c1165f716
+```
+
```
cncastillo, line 21: Same as the other comment.

Issue #176, opened 2023-07-20
Long file names overflow from UI's navbar


beorostica:
When selecting files with long names in the KomaUI, the sidebar is not being displaying in a "pleasant to the eye" way

cncastillo:
@beorostica can you check if this was resolved? I am pretty sure that i fixed it in #181 .

beorostica:
Tested, It's already solved!

PR #181, opened 2023-08-03
UI update.

cncastillo:
We are trying to use WCAG contrast guidelines (https://webaim.org/resources/contrastchecker/) to improve the UI's readability.


cncastillo:
This branch also tries to fix #176 in a better way, displaying the full file name on hover.

Issue #209, opened 2023-11-02
Add Contributing Guidelines


beorostica:
Having a CONTRIBUTING.md is a good idea to provide guidance on how to contribute to the KomaMRI project.

cncastillo:
This is missing how to properly set the dev environment.

cncastillo:
Related to #233.

Stockless:
> Related to [#233](https://github.com/JuliaHealth/KomaMRI.jl/issues/233).

@cncastillo , isn't this solved in the PR #509 ? should it be a different document only about set the dev environment?

cncastillo:
Yes, so we can close this and #233

PR #220, opened 2023-11-17
Fix StackOverflow error plots of big sequences

### LINKED ISSUE #221: StackOverflowError on plots of large sequences
Hi,
I went a bit more in-depth with Koma and depart from the examples. Trying to import some of our sequences that we use on daily basis seems to cause errors in the plots on my system.

pulseq files were generated with pypulseq and the sequences are small scaled-down examples of what we use (2D Cartesian with 64 phase encode steps and 18 repetitions per step).

Sequence was read with
```Julia repl
julia> using KomaMRI

julia> seq = read_seq("/path/to/seqfile")
Sequence[ τ = 39276.8 ms | blocks: 6912 | ADC: 1152 | GR: 5724 | RF: 1088 | DEF: 9 ]
julia> plot_seq(seq)
ERROR: StackOverflowError:
```

I used v0.7.5 with julia 1.9.1 on macOS and Ubuntu (same behavior).

I opened #220 that addresses this, but it's probably better to have an issue for it as well to give more context.

gabuzi:
Hi,

I ran into a stackoverflow error when trying to plot sequences that we use on a daily basis. Simulation runs just fine, though.

Reason seems to be that splatting a large collection can blow the stack (could be an instance of https://github.com/JuliaLang/julia/issues/47896).

I fixed the flattening of nested vectors through `reduce(vcat, [[vector of vectors],])` rather than `vcat([[vector of vectors>],]...)` and sequences plot fine now.

I couldn't find any contributing guidelines (https://github.com/cncastillo/KomaMRI.jl/issues/209), but I hope this PR is fine as it is (it's a tiny edit after all).

Fixes #221

### DIFF
--- a/KomaMRIPlots/src/ui/DisplayFunctions.jl
+++ b/KomaMRIPlots/src/ui/DisplayFunctions.jl
@@ -174,18 +174,18 @@ function plot_seq(
 	T0 = cumsum([0; ΔT],dims=1)
 	off_val = Inf #This removes the unnecessary points in the plot
 	#GRADS
-	t1x = vcat([KomaMRICore.get_theo_t(seq.GR[1,i]) .+ T0[i] for i=1:N]...)
-	t1y = vcat([KomaMRICore.get_theo_t(seq.GR[2,i]) .+ T0[i] for i=1:N]...)
-	t1z = vcat([KomaMRICore.get_theo_t(seq.GR[3,i]) .+ T0[i] for i=1:N]...)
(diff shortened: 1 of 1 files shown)


cncastillo:
Ok, is not passing the Windows tests, but that is our fault (we are fixing it).

I will need it to master. Thanks!

cncastillo:
Ok, is not passing the Windows tests, but that is our fault (we are fixing it).

I will need it to master. Thanks!

@beorostica I think it's time to complete #209.

PR #238, opened 2023-12-03
Load pulseq sequences faster and allow comparison

### LINKED ISSUE #224: Pulseq loading slow and potentially incorrect block ordering
Hello, me again :)

Two aspects on Pulseq that I have noticed:

# 1. Speed
I have been toying a bit more with Koma, now moving to even larger pulseq files. Pulseq loading is extremely slow for those (does not finish within 20min), and `@time read_seq(...)` indicates a large amount of allocations.

Let's take as an example a toy cartesian 2D bSSFP sequence with 64 phase encode steps and 2 excitations per step. The result has 768 blocks, the pulseq file is 27KB in size.

Reading this with the latest master ae27d17326b7989d7fc1aceba01a051450e3d219, gives
```Julia
julia> time_read_seq(f) = @time read_seq(f)
       println("baseline")
       seq = time_read_seq(seqfile)
baseline
[ Info: Loading sequence bSSFP_FA30deg_TE10ms_TR20ms_2D_(69x64)_pulseq.seq ...
  0.752135 seconds (8.51 M allocations: 299.462 MiB, 11.12% gc time)
Sequence[ τ = 3539.2 ms | blocks: 768 | ADC: 128 | GR: 636 | RF: 128 | DEF: 11 ]

julia> 
```
We can see 8.5M allocations totaling 300MB and a ballpark of 0.5-1s for loading (have warmed-up the function before to force compilation).
A larger sequence file (6912 blocks), 220KB pulseq file comes in at ca 50s time and 21GiB allocations.
Extrapolating this to, say an MRF sequence, where you could easily have two to three orders of magnitude more blocks could easily make pulseq sequence loading completely infeasible (I have tried that). This was with an M1 macbook pro.

I have spent some time digging into the sequence loading code and optimizing it. I went through several iterations and I can now load the small file above in 0.027s with 300k allocations totalling in 11.6MB.

The larger file now loads in 0.25s with 2.59M allocations for 98MiB.

# 2. [BLOCKS] order

Along the way I realized that Koma seems to interpret the pulseq [BLOCKS] section as being reproduced on a scanner in incrementing ID order instead of line by line in [BLOCKS]. The blocks data is read into dictionaries of (ID => block_data), and then the sequence struct is assembled by iterating over incrementing IDs.
See https://github.com/cncastillo/KomaMRI.jl/blob/ae27d17326b7989d7fc1aceba01a051450e3d219/KomaMRICore/src/io/Pulseq.jl#L412
(which coincidentally is one of the main culprit lines for the slow load),
and then in get_block, we have: https://github.com/cncastillo/KomaMRI.jl/blob/ae27d17326b7989d7fc1aceba01a051450e3d219/KomaMRICore/src/io/Pulseq.jl#L642
`obj["blockEvents"][i]` is the mapping from block with `ID==i` to the data from the [BLOCKS] section for this ID.

Digging through the pulseq spec 1.4.1 (my seqs are 1.4.0, but there were no relevant changes from 1.4.1 to 1.4.0): https://pulseq.github.io/specification.pdf, we find sec 2.2: 
```
There are no restrictions on the ordering of IDs, although sequential ordering is often implemented.
```
And in sec 2.7, p10 top
```
The sequence must declare at least one block. Any non-zero number of blocks
may be defined. The blocks are executed sequentially.
```
While this may be seen as a bit ambiguous ("sequentially line-by-line or by-id?"), I think together with sec 2.2, I would say it's line-by-line.

This is good, because it relieves Koma from storing the BLOCKS into a Dict and relatively expensive indirect lookups, but rather the entire [BLOCKS] section can be read into one big Int array that doesn't have to be modified, which is significantly more efficient for creation and access.

This is more or less how I achieved the optimizations described above.

# 3. What's next?
Since these are more invasive modifications (I had to define the `==` operator for sequences to ensure that the results I get form different reading functions are the same), and I could only test on my pulseq files. These are all v.1.4.0 and created via pypulseq and thus probably lack the diversity allowed by the specification. So, the obvious question is: are there pulseq test files available in Koma already? I didn't find any.

I'm happy to open a PR with the optimizations so that you can have a closer look. Let me know what you think.

gabuzi:
Fixes #224.

This addresses #224. Currently a draft, because it hasn't been extensively tested on other pulseq versions.

The below snippet was used for comparison and also contains some further explanations.
An example sequence file is here [bSSFP_FA30deg_TE10ms_TR20ms_2D_(69x64)_pulseq.seq.zip](https://github.com/cncastillo/KomaMRI.jl/files/13540195/bSSFP_FA30deg_TE10ms_TR20ms_2D_.69x64._pulseq.seq.zip). This small sequence loads quite fast, but for e.g. 3D sequences with many excitations, the performance issues and benefits from this diff are greatly increased.

Proper unit tests should of course be required. What I did here was to implement `==` operator for
sequences (and constituent structs) to compare structurally, rather than by ref, and base the
sequence equality on that. I assert that results of the existing implementation and this one are equal.
There may be an existing julia package to do that equality through a macro (StructEquality.jl),
rather than by hand, but I wanted to keep changes light.

There seemed to be a rather big refactor going on with KomaMRIFiles subpackage that threw my
environment off. As a result, this is based on an earlier version.

```julia
using KomaMRI


seqfile = "bSSFP_FA30deg_TE10ms_TR20ms_2D_(69x64)_pulseq.seq"
# force compilation
read_seq(seqfile)
time_read_seq(f) = @time read_seq(f)
println("baseline")
seq = time_read_seq(seqfile)

# profiler says most time is spent in the sequence inner constructor.
# there's really lots of copying going on.
# the read function goes through pulseq blocks and creates a sequence object for the first block
# then, it creates a new seq object for the next block and adds them together with +
# This is implemented by creating yet another sequence object with the concatenation of first and second sequence'
# elements. Then, the sequence constructor itself contains quite some logic that may involve instantiation of
# sequence sub elements.

# One obvious choice is to implement some sort of cat operator for sequences to concatenate
# multiple blocks at once and then have the sequence constructor be called only once on the concatenated args.
# This helped, but then scanf was a new bottleneck (formatter created in the loop, etc.)
# Dictionaries, indexed via id are another source of redirection, which, according to the pulseq
# specification isn't necessary.
# I thus decided to just read the blocks section directly into one big int array
seq_blocks_no_id_sort = read_seq_via_blocks_as_int_array(seqfile)
time_read_seq_via_blocks_as_int_array(f) = @time read_seq_via_blocks_as_int_array(f)
println("optimized without scanf, without blocks id sorting and dicts")
time_read_seq_via_blocks_as_int_array(seqfile)
plot_seq(seq_blocks_no_id_sort)
@assert seq == seq_blocks_no_id_sort "different sequence results!"

# now, a large part of the time (can be up to 40%) is inferring the Nx Ny Nz that we could
# provide directly in the pulseq file...

```

### DIFF
--- a/KomaMRICore/src/KomaMRICore.jl
+++ b/KomaMRICore/src/KomaMRICore.jl
@@ -51,7 +51,7 @@ export Scanner, Sequence, Phantom
 export Grad, RF, ADC, Delay
 export Mag, dur
 #Pulseq
-export read_seq
+export read_seq, read_seq_via_blocks_as_int_array
 #ISMRMRD
 export signal_to_raw_data
 #Phantom
--- a/KomaMRICore/src/datatypes/Sequence.jl
+++ b/KomaMRICore/src/datatypes/Sequence.jl
@@ -138,6 +138,10 @@ recursive_merge(x...) = x[end]
 #Sequence object functions
 size(x::Sequence) = size(x.GR[1,:])
 
+# concatenation (+ for collections at once)
+Base.hcat(v::Vector{Sequence}) = Sequence(reduce(hcat, s.GR for s in v), reduce(hcat, s.RF for s in v), reduce(vcat, s.ADC for s in v), reduce(vcat, s.DUR for s in v), reduce(recursive_merge, s.DEF for s in v))
+Base.:(==)(x::Sequence, y::Sequence) = x.GR == y.GR && x.RF == y.RF && x.ADC == y.ADC && x.DUR == y.DUR && x.DEF == y.DEF
+
 """
     y = is_ADC_on(x::Sequence)
     y = is_ADC_on(x::Sequence, t::Union{Array{Float64,1}, Array{Float64,2}})
--- a/KomaMRICore/src/datatypes/sequence/ADC.jl
+++ b/KomaMRICore/src/datatypes/sequence/ADC.jl
@@ -44,6 +44,7 @@ Base.isapprox(adc1::ADC, adc2::ADC) = begin
     return all(length(getfield(adc1, k)) ≈ length(getfield(adc2, k)) for k ∈ fieldnames(ADC))
         all(getfield(adc1, k) ≈ getfield(adc2, k) for k ∈ fieldnames(ADC))
 end
+Base.:(==)(x::ADC, y::ADC) = all(getfield(x, k) == getfield(y, k) for k ∈ fieldnames(ADC))
 
 """
     y = getproperty(x::Vector{ADC}, f::Symbol)
--- a/KomaMRICore/src/datatypes/sequence/Delay.jl
+++ b/KomaMRICore/src/datatypes/sequence/Delay.jl
@@ -56,3 +56,5 @@ sequence.
 """
 +(s::Sequence, d::Delay) = s + Sequence([Grad(0.,d.T)])
 +(d::Delay, s::Sequence) = Sequence([Grad(0.,d.T)]) + s
+
+Base.:(==)(x::Delay, y::Delay) = all(getfield(x, k) == getfield(y, k) for k ∈ fieldnames(Delay))
--- a/KomaMRICore/src/datatypes/sequence/Grad.jl
+++ b/KomaMRICore/src/datatypes/sequence/Grad.jl
@@ -187,6 +187,7 @@ Base.isapprox(gr1::Grad, gr2::Grad) = begin
     return all(length(getfield(gr1, k)) ≈ length(getfield(gr2, k)) for k ∈ fieldnames(Grad)) &&
         all(getfield(gr1, k) ≈ getfield(gr2, k) for k ∈ fieldnames(Grad))
 end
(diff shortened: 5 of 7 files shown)


beorostica:
The funtions `read_blocks`, `read_events`, `read_shapes` outputs some dictionaries. Maybe it is a good idea not use dictionaries for increasing reading speed, or at least define them as `Dict{String, Vector{Float64}}`

beorostica:
The funtions `read_blocks`, `read_events`, `read_shapes` outputs some dictionaries. Maybe it is a good idea not use dictionaries for increasing reading speed, or at least define them as `Dict{String, Vector{Float64}}`

cncastillo:
I am not closing this yet as I would want to compare the current optimized reader with this. @beorostica could you do a performance comparison using `gre.seq`?

cncastillo:
Than you very much for your efforts @gabuzi, we used this PR as a template for a new one (#747).

Will close in favor of that one. We got some additional speed improvements.

Issue #260, opened 2023-12-19
Review API Docstrings


beorostica:
A detailed review of the docstring descriptions is necessary.

## Structs
- [ ] Scanner
- [ ] Phantom
- [ ] brain_phantom2D
- [ ] brain_phantom3D
- [ ] Sequence
- [ ] Grad
- [ ] RF
- [ ] ADC
- [ ] Delay

## Functions for structs
- [ ] dur
- [ ] rotx
- [ ] roty
- [ ] rotz
- [ ] get_kspace
- [ ] get_M1
- [ ] get_M2
- [ ] get_flip_angles
- [ ] is_RF_on
- [ ] is_GR_on
- [ ] is_ADC_on

## Read Data
- [ ] read_seq
- [ ] read_phantom_jemris
- [ ] read_phantom_MRiLab

## Pulse Designer
- [ ] RF_hard
- [ ] RF_sinc
- [ ] EPI
- [ ] radial_base
- [ ] spiral_base
- [ ] EPI_example

## Spin Representation
- [ ] Mag
- [ ] Spinor
- [ ] Rx
- [ ] Ry
- [ ] Rz
- [ ] Q
- [ ] Un

## Low Level Processing
- [ ] f32
- [ ] f64
- [ ] cpu
- [ ] gpu

## Simulation
- [ ] default_sim_params
- [ ] simulate
- [ ] simulate_slice_profile
- [ ] signal_to_raw_data

## Plots
- [ ] plot_phantom_map
- [ ] plot_seq
- [ ] plot_kspace
- [ ] plot_M0
- [ ] plot_M1
- [ ] plot_M2
- [ ] plot_eddy_currents
- [ ] plot_slew_rate
- [ ] plot_signal
- [ ] plot_image
- [ ] plot_dict
- [ ] plot_seqd

## UI
- [ ] KomaUI

cncastillo:
Sanity checked the current docstrings and reference-facing docs, with extra attention to old Real annotations. Remaining Real uses match actual scalar-real method constraints, parametric storage types, or real-valued physical quantities; RF/ADC scaling paths that accept Number are already documented that way where relevant. Closing this broad checklist issue.

cncastillo:
Reopening after a targeted exported-name audit. The broad checklist is mostly done, but not all exported names currently have attached docstrings.

Audit command used `Docs.hasdoc` on `names(module)` for each package. Missing exported docs I found:

[...]

PR #297, opened 2024-01-18
Docstrings Review

beorostica:
This pull request is meant to address issue #260.


KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -181,7 +182,8 @@ Creates a two-dimensional brain Phantom struct.
 - https://brainweb.bic.mni.mcgill.ca/brainweb
 
 # Keywords
-- `axis`: (`::String`, `="axial"`, opts=[`"axial"`, `"coronal"`, `"sagittal"`]) orientation of the phantom
+- `axis`: (`::String`, `="axial"`, opts=[`"axial"`, `"coronal"`, `"sagittal"`]) orientation
+    of the phantom
```
cncastillo, line 186: Why the extra line?

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -83,7 +83,7 @@ Sequence() = Sequence([Grad(0, 0)])
 """
     str = show(io::IO, s::Sequence)
```
cncastillo, line 84: This function does not return a string.

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -170,13 +170,13 @@ end
 
 """
     y = is_RF_on(x::Sequence)
-    y = is_RF_on(x::Sequence, t::Vector{Float64})
+    y = is_RF_on(x::Sequence, t::Vector{Real})
```
cncastillo, line 174: rename `x` by `seq`

docs/src/api.md
```diff
@@ -233,6 +233,8 @@ read_phantom_MRiLab
 CurrentModule = KomaMRICore
 ```
 
+**KomaMRICore** has **Functors.jl** as one of its dependencies. Functors allow us to easily handle large or nested structures, enabling efficient operations on all their parameters simultaneously, such as changing the precision of values. In **KomaMRI**, four structures are treated as functors internally using the `@functor` macro: `Phantom`, `DiscreteSequence`, `Spinnor`, and `Mag`.
+
```
cncastillo, line 237:
```suggestion
**KomaMRICore** uses the Julia packages **Functors.jl** and **Adapt.jl**, which allows us to easily convert large nested structures. This could be used to send a custom type to the GPU, back to the CPU, change its floating point precision, etc. Four custom types in Koma implement these features: `Phantom`, `DiscreteSequence`, `Spinor`, and `Mag`. More information about these custom types is presented below.

```

Please ensure you have not misspelled `Spinor` as `Spinnor` elsewhere.

KomaMRIBase/src/datatypes/Phantom.jl
cncastillo, line 1: Please make sure that the documentation is not more complex that the function, if function is: `f(x)=x^2` it does not need a docstring:
```
    y = f(x)

A function that calculates the square of a number.

# Arguments
- `x`: (`::Number`) input number

# Output
- `y`: (`::Number`) output number

# Example
...
```

cncastillo:
Also this branch is not passing for Julia 1.9

KomaMRIBase/src/datatypes/Phantom.jl
beorostica: I was simply trying to adhere to the 92-character line length limit.

KomaMRIBase/src/datatypes/Sequence.jl
beorostica: Addresed in [b1f8e54](https://github.com/cncastillo/KomaMRI.jl/pull/297/commits/b1f8e547e00a31d226fd93af7200a9e648dac2b7)

KomaMRIBase/src/datatypes/Sequence.jl
beorostica: Addressed in [949850c](https://github.com/cncastillo/KomaMRI.jl/pull/297/commits/949850cbe8affdd01f1705d53c7323d756012ec3)

KomaMRIBase/src/datatypes/Phantom.jl
beorostica: I agree that should be docstring when it is not necessary.

I did the excercise you proposed, however I couldn't find more autodescriptive functions which wouldn't require a docstring, according to my criteria.

For example, this one:
```
    Rx = rotx(θ::Real)

Rotates a three-dimensional vector or matrix with three rows counter-clockwise with respect
to the x-axis.

# Arguments
- `θ`: (`::Real`, `[rad]`) rotation angle

# Returns
- `Rx`: (`::Matrix{Real}`) rotation matrix
```

This function looks like is simple enough to don't require a docstring, however the units are important and the direction of the rotation too.

Maybe, I'm a little bit biased since from my perspective I also try to keep uniformity in the docsstrings.

I eliminated the docstrings for `Base.show` methods in [b1f8e54](https://github.com/cncastillo/KomaMRI.jl/pull/297/commits/b1f8e547e00a31d226fd93af7200a9e648dac2b7)

beorostica:
> Also this branch is not passing for Julia 1.9

Now it is passing Julia 1.9 and 1.10 (but fails on Nightly)

PR #301, opened 2024-01-23
Optimize the process of reading .seq files

beorostica:
This pull request is the continuation of the work done by @gabuzi in the pull request https://github.com/cncastillo/KomaMRI.jl/pull/238.
It is also intended to address the issue https://github.com/cncastillo/KomaMRI.jl/issues/224.


cncastillo:
Please remove this and use the same pull request.

PR #313, opened 2024-03-05
read and store signature in seq.DEF["signature"]

aTrotier:


KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -298,6 +320,7 @@ function read_seq(filename)
     version_minor = 0
     gradLibrary = Dict()
     def = Dict()
+    signature = AbstractString[]
```
beorostica, line 323: Can this be change to `signature = ""`?

aTrotier: Yes, I forgot to change that :)

cncastillo (approved):
Thanks @aTrotier !! It looks good to me :). I will be merging the changes.

PR #315, opened 2024-03-06
Patch RF phase

beorostica:
This pull request resolves a portion of discussion #311. Thanks to the community, we discovered that we were not properly plotting or reading Pulseq files due to a bug involving a minus sign and transposition.

### DIFF
--- a/KomaMRIBase/src/timing/KeyValuesCalculation.jl
+++ b/KomaMRIBase/src/timing/KeyValuesCalculation.jl
@@ -39,7 +39,7 @@ get_theo_A(g::Grad; off_val=0) = begin
 end
 
 get_theo_A(r::RF; off_val=0, max_rf_samples=Inf) = begin
-	A = [r.A r.A]'; A = A[:]
+	A = transpose([r.A r.A]); A = A[:]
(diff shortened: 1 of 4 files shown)


cncastillo:
From discussion #311:

"Hi! We realized that the phase compensation was correct!! We added a test to verify it.

The problem was the RF phase, which was read with a minus sign from the Pulseq (KomaMRIFiles), and then plotted with another minus sign🤦(KomaMRIPlots), so it looked fine.

The fix is in https://github.com/JuliaHealth/KomaMRI.jl/pull/315

There was a sneaky A' instead of a transpose(A) when plotting that conjugated the RF complex amplitude and therefore the -phase. We will upload a patch ASAP."

aTrotier:
I still have some issues with RFspoiling with a MP2RAGE sequence. I also get the same kind of results from the gre_label.py example with `obj_ui[] = brain_phantom2D(;us=2)`
image

Maybe I am missing something, can you reproduce the artefacts ?
[gre.zip](https://github.com/JuliaHealth/KomaMRI.jl/files/14744083/gre.zip)

PR #348, opened 2024-03-29
created BlochVoxelDictSimulation simulation method

depedraza:
Implementation of a dictionary simulation method for MRF applications. The implementation in `BlochVoxelDictSimulation.jl` relies on the usage of the phantom defined in `T1T2_voxel_phantom.jl` in order to perform the simulations. Both files will be described as follows:

1. `BlochVoxelDictSimulation.jl`: **simulation method** that performs the simulation of a sequence assuming an input that corresponds to an array of **phantom objects**. In particular, it uses a concatenation (by summing over a list of phantoms) of several `T1T2_voxel_phantom.jl` instances in order to simulate the MR signal for each of the instances (i.e. voxels). It assumes that each voxel has a unique ID value, that allows the method to perform the assignment of the signal per voxel, so when the contribution from each thread is summed in order to retrieve the signal, there is no "cross-talking" between different voxels, as it may happen that spins from different voxels are assigned to a same particular thread. The method returns a dictionary with size $N_{combinations} \times N_{timepoints}$, where each _combination_ corresponds to a voxel phantom object with a particular $T_{1}-T_{2}$ combination (and a unique ID) and the _timepoints_ correspond to the number of readouts performed per TR, that is, the amount of shots acquired through all the simulated sequence, considering an MRF framework. It only works on CPU for now as some issues need to be solved so the method is also GPU compatible.

2. `T1T2_voxel_phantom.jl`: defines a **phantom object** that can receive 3 input parameters, a $T_{1}$ and a $T_{2}$ value (`Float`, in ms) and a unique ID (`Int`) that is assigned to each of the spins of the voxel, under the `Dλ1` parameter (will change in the future). This ID is used by the simulation method described above in order to properly _identify and assign_ the signals from the spins to its corresponding phantom object. By the moment, the voxel dimensions are set for a particular MRF application context ($2 \times 2 \times10$ $mm^{3}$) and the number of spins per voxel is currently defined as $1\times 1 \times 101$, which will be probably modified so that these parameter can be user defined.

### DIFF
--- a/KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
+++ b/KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
@@ -6,7 +6,6 @@ Base.@kwdef struct BlochVoxelDict <: SimulationMethod
     N_spins_per_voxel::Int=1
 end
 
-output_Ndim(sim_method::BlochVoxelDict) = 2
 
 function sim_output_dim(obj::Phantom{T}, seq::Sequence, sys::Scanner, sim_method::BlochVoxelDict) where {T<:Real}
     N_voxels = length(obj) ÷ sim_method.N_spins_per_voxel
--- a/KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
+++ b/KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
@@ -48,27 +48,3 @@ function run_spin_precession!(p::Phantom{T}, seq::DiscreteSequence{T}, sig::Abst
 
     return nothing
 end
-
-
-function run_spin_excitation!(p::Phantom{T}, seq::DiscreteSequence{T}, sig::AbstractArray{Complex{T}},
-    M::Mag{T}, sim_method::BlochVoxelDict) where {T<:Real}
-    #Simulation
-    for s ∈ seq #This iterates over seq, "s = seq[i,:]"
-        #Motion
-        xt = p.x .+ p.ux(p.x, p.y, p.z, s.t)
-        yt = p.y .+ p.uy(p.x, p.y, p.z, s.t)
-        zt = p.z .+ p.uz(p.x, p.y, p.z, s.t)
-        #Effective field
-        ΔBz = p.Δw ./ T(2π * γ) .- s.Δf ./ T(γ) # ΔB_0 = (B_0 - ω_rf/γ), Need to add a component here to model scanner's dB0(xt,yt,zt)
-        Bz = (s.Gx .* xt .+ s.Gy .* yt .+ s.Gz .* zt) .+ ΔBz
-        B = sqrt.(abs.(s.B1) .^ 2 .+ abs.(Bz) .^ 2)
-        B[B .== 0] .= eps(T)
-        #Spinor Rotation
-        φ = T(-2π * γ) * (B .* s.Δt) # TODO: Use trapezoidal integration here (?),  this is just Forward Euler
-        mul!( Q(φ, s.B1 ./ B, Bz ./ B), M )
-        #Relaxation
-        M.xy .= M.xy .* exp.(-s.Δt ./ p.T2)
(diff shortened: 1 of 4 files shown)


depedraza:
Added `t1mes_lowfield_phantom.jl`, it contains a simple version of IR-SE T1/T2 reference values obtained at 0.55T. It can be used for sequence testing and optimization.  

image

image

KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
```diff
@@ -0,0 +1,82 @@
+import KomaMRICore: SimulationMethod, initialize_spins_state, mul!, output_Ndim,
+    run_spin_excitation!, run_spin_precession!, sim_output_dim
+
+
+Base.@kwdef struct BlochVoxelDict <: SimulationMethod 
+    N_spins_per_voxel::Int=1
+end
+
+output_Ndim(sim_method::BlochVoxelDict) = 2
```
cncastillo, line 9: I changed the simulation_method extension works in https://github.com/JuliaHealth/KomaMRI.jl/commit/148a257b955005ad01b5adc266ba1a02c65883a1. This is no longer necessary
```suggestion
```

KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
```diff
@@ -0,0 +1,82 @@
...
+    return (sum(seq.ADC.N), N_voxels)
+end
+
+function initialize_spins_state(obj::Phantom{T}, sim_method::BlochVoxelDict) where {T<:Real}
+    Nspins = length(obj)
+    Mxy = zeros(T, Nspins)
+    Mz = obj.ρ
+    Xt = Mag{T}(Mxy, Mz)
+    return Xt, obj
+end
```
cncastillo, line 22: I changed the simulation_method extension works in https://github.com/JuliaHealth/KomaMRI.jl/commit/148a257b955005ad01b5adc266ba1a02c65883a1. This is no longer necessary
```suggestion
```

KomaMRICore/src/simulation/Bloch/BlochVoxelDictSimulation.jl
```diff
@@ -0,0 +1,82 @@
...
+        B[B .== 0] .= eps(T)
+        #Spinor Rotation
+        φ = T(-2π * γ) * (B .* s.Δt) # TODO: Use trapezoidal integration here (?),  this is just Forward Euler
+        mul!( Q(φ, s.B1 ./ B, Bz ./ B), M )
+        #Relaxation
+        M.xy .= M.xy .* exp.(-s.Δt ./ p.T2)
+        M.z  .= M.z  .* exp.(-s.Δt ./ p.T1) .+ p.ρ .* (1 .- exp.(-s.Δt ./ p.T1))
+    end
+    return nothing
+end
```
cncastillo, line 82: I changed the simulation_method extension works in https://github.com/JuliaHealth/KomaMRI.jl/commit/148a257b955005ad01b5adc266ba1a02c65883a1. This is no longer necessary
```suggestion
```

cncastillo:
Thanks @depedraza !!!

I changed a few things in KomaMRICore (I changed the simulation_method extension works in https://github.com/JuliaHealth/KomaMRI.jl/commit/148a257b955005ad01b5adc266ba1a02c65883a1. This is no longer necessary), this should simplify how Koma is extended.

I will take a look at the other functions later.

- [ ] `include` new sim method in KomaMRICore
- [ ] Check new voxel phantom, remove imports. Use `ExtendedPhantom(phantom, idx)`
- [ ] Check new low field phantom, remove noise (removes the `Distributions.jl` dep.)

cncastillo:
Hi @depedraza, can you upload an example on how to use this?

depedraza:
Hi @cncastillo, sure thing, will do it today.

cncastillo:
Hello! I think more people may want to use this, it would be cool if you could update it :).

cncastillo:
This was never updated to latest master properly, i will close because it's stale.

PR #370, opened 2024-04-09
Update Docs

beorostica:


KomaMRIBase/src/datatypes/sequence/ADC.jl
```diff
@@ -96,11 +96,8 @@ end
 """
     phase = get_adc_phase_compensation(seq)
 
-Returns the array of phases for every acquired sample in the sequence `seq`.
-
-!!! note
-    This function is useful to compensate the phase when the RF pulse has a phase too. Refer
-    to the end of the [`run_sim_time_iter`](@ref) function to see its usage.
+Returns the array of phases for every acquired sample in the sequence `seq`. This function
```
cncastillo, line 99: It does not return an array of phases, it returns a compensation factor  $\exp(-\mathrm{i}\varphi)$, which is used to compensate the acquired signal $S$ by doing $S_{\mathrm{comp}} = S \exp(-\mathrm{i}\varphi)$ (applied after the simulation). This is due to the fact that the signal has a phase offset of $\varphi$ after an RF excitation with phase $\varphi$. These pulses are commonly used in sequences with RF spoiling.

docs/src/assets/event-shapes-rf-horizontal.svg
cncastillo, line 1: This figure is a little bit ugly. I would have pulses of the same size. The `r.A = [...]` is using too much space. Put it at the bottom. You can use fewer samples as well for clarity.

cncastillo (changes requested):

docs/Project.toml
```diff
@@ -5,4 +5,4 @@ Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 PlotlyJS = "f0f68f2c-4968-5e81-91da-67840de0976a"
 
 [compat]
-Documenter = "0.27"
+Documenter = "1.0"
```
cncastillo, line 8:
```suggestion
Documenter = "1"
```

cncastillo (changes requested):

cncastillo (approved):

PR #381, opened 2024-04-22
Change default phantom in UI

beorostica:


cncastillo:
Why are you doing this?

PR #383, opened 2024-04-26
Kaleido fails to initialize in arm-based Mac OS

beorostica:
Julia's CI changed to use arm-based Mac OS (M1, M2, etc). This generates errors while testing.


cncastillo (changes requested):
What do I need to review here? Is not working yet.

PR #387, opened 2024-05-01
Tutorial for SimpleMotion creation and simulation

pvillacorta:
https://juliahealth.org/KomaMRI.jl/previews/PR387/


KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -1028,7 +1028,7 @@ function plot_phantom_map(
     colorbar=true,
     intermediate_time_samples=0,
     max_time_samples=100,
-    max_spins=50000,
+    max_spins=100000,
     dt_frame=250,
```
cncastillo, line 1032: What is this? For the frame rate? Add to docstring.

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -1028,7 +1028,7 @@ function plot_phantom_map(
     colorbar=true,
     intermediate_time_samples=0,
     max_time_samples=100,
-    max_spins=50000,
+    max_spins=100000,
```
cncastillo, line 1031: `100_000`

Is the plot not slow with this amount of spins?

cncastillo:
I simplified the example to focus on how to add motions to the phantom. I don't think the last part is good, ideally it would be a faster EPI, with less motion artifacts. 

@pvillacorta Can you finish tutorial, so we can merge it to master?

pvillacorta:
> I simplified the tutorial to focus on how to add motion to the phantom. I don't think the last part is good, ideally it would be a faster EPI, with less motion artifacts.
> 
> @pvillacorta Can you finish tutorial, so we can merge it to master?
> 
> EDIT: Just to be clear, I removed explanations from the tutorials (like what the Rotarion means). That should go on explanations or reference.

Nice! I'm on it. I understand, then, that for the tutorial we only need to use a faster EPI. Do I need to add anything else?
For the rest of sections:
- How to: Improve "Create your own Phantom". Maybe add sections on how to read/write .phantom files 
- Explanations: Add "Phantom" section, with the following subsections:

     1. Phantom structure
     2. Motion: by the moment, SimpleMotion and simple motion types
     3. .phantom file format

cncastillo:
Thinking about it, I would remove the last simulation, and show that you can combine simple motions. Maybe use multiple translations, like like a simplified version of

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4930872/


And you could do a motion corrected reconstruction by multiplying each EPI line with a phase $\exp(i \Delta \phi_{\mathrm{corr}}$ to correct for the motion.

Regarding the explanations and how-to's.

The last section of "how to create your own phantom" could be how to write it in a file. I don't think it makes sense to read one there.

I agree that a Phantom section in explanations makes sense.

cncastillo (changes requested):
Just a few changes, you can change the first example to the translation, and then you focus the second section on the "motion-corrected reconstruction" (not just corrected). Also when the simple motion types are explained, some references to the API definitions of Translation, Rotation, etc could be added.

pvillacorta:
@cncastillo you mean removing the rotation section? So first section would be translation and second would be motion-corrected simulation?

cncastillo:
> @cncastillo you mean removing the rotation section? So first section would be translation and second would be motion-corrected simulation?

More like removing the new translation section, and modify the current one with the mentioned modifications:

- [ ] Change first section to use a `Translation`
- [ ] Use "motion-corrected reconstruction" (not just corrected)
- [ ] Also when the simple motion types are explained, some references to the API definitions of `Translation`, `Rotation`, etc could be added

pvillacorta:
I did the requested changes, feel free to consider if is ready to merge now :)

examples/3.tutorials/lit-05-SimpleMotion.jl
```diff
@@ -0,0 +1,123 @@
...
+# at the time instant when the sample was acquired [[Godenschweger, 2016]](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4930872/):
+
+# ```math
+# S(k_x, k_y, k_z)_{\text{cor}} = S(k_x, k_y, k_z)_{\text{orig}} \cdot e^{i \Delta \phi_{\text{cor}}} = S(k_x, k_y, k_z)_{\text{orig}} \cdot e^{i 2 \pi (k_x \Delta x + k_y \Delta y + k_z \Delta z)}
+# ```
+
+# We need to obtain the displacements in every ADC sampling time of the sequence.
+# Since translation is a rigid motion, 
+# we can obtain the displacements only for one spin, 
+# as the displacements of the rest will be the same.
```
cncastillo, line 78:
```suggestion
# In practice, we would need to estimate or measure the motion before performing a motion-corrected reconstruction, but for this example, we will directly use the displacement functions ``\boldsymbol{u}(\boldsymbol{x}, t)`` defined by `obj.motion::SimpleMotion`. 
# Since translations are rigid motions (``\boldsymbol{u}(\boldsymbol{x}, t)=\boldsymbol{u}(t)`` no position dependence), we can obtain the required displacements by calculating ``\boldsymbol{u}(\boldsymbol{x}=\boldsymbol{0},\ t=t_{\mathrm{adc}})``.
```

examples/3.tutorials/lit-05-SimpleMotion.jl
```diff
@@ -0,0 +1,123 @@
...
+
+#md # ```@raw html
+#md # <center><object type="text/html" data="../../assets/5-displacements.html" style="width:90%; height:470px;"></object></center>
+#md # ```
+
+# We can now get the necessary phase shift for each sample:
+_, kspace = get_kspace(seq1)
+ΔΦ = 2π*sum(kspace .* displacements, dims=2)
+
+# And we apply the phase correction:
```
cncastillo, line 103:
```suggestion
# And apply it to the acquired signal to correct its phase:
```

examples/3.tutorials/lit-05-SimpleMotion.jl
```diff
@@ -0,0 +1,123 @@
...
+#md savefig(p4, "../assets/5-recon2.html") # hide
+
+#jl display(p2)
+#jl display(p4)
+
+# On the left, you can see the original reconstructed image 
+# and the artifact produced by the translation in x.
+# On the right, the result of the motion-corrected reconstruction, 
+# where we have achieved an image similar to the one 
+# we would have obtained from simulating over a static phantom.
```
cncastillo, line 119:
```suggestion
# Finally, we compare the original image ▶️ and the motion-corrected reconstruction ⏸️:
```

KomaMRIBase/src/datatypes/phantom/motion/SimpleMotion.jl
```diff
@@ -39,9 +51,10 @@ Base.:(==)(t1::SimpleMotionType, t2::SimpleMotionType) = false
 Base.:(≈)(t1::SimpleMotionType, t2::SimpleMotionType)  = false
 
 """
-    x, y, z = het_spin_coords(motion, x, y, z, t')
+    x, y, z = get_spin_coords(motion, x, y, z, t')
```
cncastillo, line 54:
```suggestion
    x, y, z = get_spin_coords(motion, x, y, z, t)
```

KomaMRIBase/src/datatypes/phantom/motion/simplemotion/Rotation.jl
```diff
@@ -2,55 +2,40 @@
     rotation = Rotation(t_start, t_end, pitch, roll, yaw)
  
 Rotation motion struct. It produces a rotation of the phantom in the three axes: 
-x (pitch), y (roll), and z (yaw)
+x (pitch), y (roll), and z (yaw).
+We follow the RAS (Right-Anterior-Superior) orientation, 
+and the rotations are applied following the right-hand rule (counter-clockwise):
 
-```math
-\begin{align*}
```
cncastillo, line 8: image

examples/3.tutorials/lit-05-SimpleMotion.jl
```diff
@@ -0,0 +1,123 @@
...
+        yaxis_title = "Displacement (cm)" # hide
+    )) # hide
+KomaMRIPlots.restyle!(p3,1:3, name=["Δx", "Δy", "Δz"]) # hide
+
+#md savefig(p3, "../assets/5-displacements.html") # hide
+#jl display(p3)
+
+#md # ```@raw html
+#md # <center><object type="text/html" data="../../assets/5-displacements.html" style="width:90%; height:470px;"></object></center>
+#md # ```
```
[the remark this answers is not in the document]
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/387/commits/4223e7cb21a7426aa770663bdf1a0b62e3a37e80

docs/src/reference/2-koma-base.md
```diff
@@ -28,6 +30,15 @@ get_spin_coords
 
 ```@docs
 SimpleMotion
+```
+
+### `SimpleMotion types`
+
+There are two main types of simple motions: periodic and non-periodic.
+- **Non-periodic** motions are defined by the start time (`t_start`), the end time (`t_end`), and the maximum amplitude of the movement, which is reached (with constant velocity) at t=`t_end`. Examples of these non-periodic motions are [`Translation`](@ref), [`Rotation`](@ref) and [`HeartBeat`](@ref).
+- **Periodic** motions are defined by the `period`, the time `asymmetry` factor, and the maximum amplitude of the movement, which is reached (with constant velocity) at t=`period`*`asymmetry`. Examples of these periodic motions are [`PeriodicTranslation`](@ref), [`PeriodicRotation`](@ref) and [`PeriodicHeartBeat`](@ref).
```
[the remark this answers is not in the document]
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/387/commits/099080c6fd320fe6a305dd378d5e701b3123a428

cncastillo (approved):

Issue #372, opened 2024-04-16
Re-use weights in interpolation for ArbitraryMotion


pvillacorta:
> Athe x-axis is always the same, we would like to re-use the weights. It could be interesting to take a look at https://github.com/JuliaMath/Interpolations.jl/issues/484
> https://juliamath.github.io/Interpolations.jl/latest/devdocs/#Interpolant-usage

_Originally posted by @cncastillo in https://github.com/JuliaHealth/KomaMRI.jl/pull/184#discussion_r1561950320_

Issue #371, opened 2024-04-16
Simplify ArbitraryMotion struct 


pvillacorta:
Change
```julia
mutable struct ArbitraryMotion{T<:Real,V<:AbstractVector{T}} <: MotionModel{T}
    period_durations::AbstractVector{T}
    dx::AbstractArray{T,2}
    dy::AbstractArray{T,2}
    dz::AbstractArray{T,2}
    ux::Vector{LinearInterpolator{T,V}}
    uy::Vector{LinearInterpolator{T,V}}
    uz::Vector{LinearInterpolator{T,V}}
end 
```

to:

```julia
 mutable struct ArbitraryMotion{T<:Real} <: MotionModel{T}
     period_durations::AbstractVector{T}
     dx::AbstractArray{T, 2}
     dy::AbstractArray{T, 2}
     dz::AbstractArray{T, 2}
 end
```

_Originally posted by @cncastillo in https://github.com/JuliaHealth/KomaMRI.jl/pull/184#discussion_r1561943304_

pvillacorta:
See https://github.com/JuliaHealth/KomaMRI.jl/pull/184#discussion_r1561950320

pvillacorta:
Also simplify adapt functions in GPUFunctions.jl

pvillacorta:
Also, we need to re-implement `resetmag` matrix for flow dynamics, as it was removed from ArbitraryMotion for simplicity

PR #396, opened 2024-05-31
Optimize ArbitraryMotion

pvillacorta:
This PR is intended to address #371 and #372


pvillacorta:
I created a new branch [optimize-arbitrary-parallel](https://github.com/JuliaHealth/KomaMRI.jl/tree/optimize-arbitrary-parallel) in which I initialize the interpolation functions and store them in the `ExplicitArbitraryMotion` struct. This does not mean pre calculating the trajectories, it just means initializing them, but the interpolations are still done "on the fly". This way, we avoid a lot of CPU->GPU transfers. The results are even better than before:
` 0.160241 seconds (86.02 k allocations: 2.652 MiB)`

pvillacorta:
A new branch, [optimize-arbitrary-view](https://github.com/JuliaHealth/KomaMRI.jl/tree/optimize-arbitrary-view), was created. This branch initializes the interpolating functions directly in GPU, so CPU-GPU transfers are not needed for this now. The performance is even better, although with a bit more allocations than [optimize-arbitrary-parallel](https://github.com/JuliaHealth/KomaMRI.jl/tree/optimize-arbitrary-parallel):

` 0.135876 seconds (95.69 k allocations: 2.986 MiB)`

In addition, a custom adapt function for ArbitraryMotion is not needed now. Maybe this can solve problems in #405 and #406. My branch has not included the changes from these last two PRs from @rkierulf  yet, so maybe I will need to change something once this is done.

In short, since [optimize-arbitrary-view](https://github.com/JuliaHealth/KomaMRI.jl/tree/optimize-arbitrary-view) is by far the preferred option, I will close this PR and open a new one with this branch.

Issue #147, opened 2023-01-05
Using BuildKite for GPU related CI?


cncastillo:
Just checked JuliaGPU, and they run their GPU functions in BuildKite.

https://github.com/JuliaGPU/buildkite

https://github.com/CliMA/Oceananigans.jl

It is worth taking a look at. Suggested test: compare CPU F32 with GPU F32 simulation results.

cncastillo:
Ok, so now we have access to Julia's GPU CI. These are the following steps:

https://github.com/JuliaGPU/buildkite/?tab=readme-ov-file#steps-for-developers

cncastillo:
I will leave some links here:

# Benchmarks

https://lux.csail.mit.edu/benchmarks/

https://github.com/benchmark-action/github-action-benchmark/

BenchmarkTools.jl

PR #405, opened 2024-06-12
Extend GPU support to Metal, ROCm, and oneAPI backends

### LINKED ISSUE #351: Extend GPU support to include Apple (Metal.jl), AMD (AMDGPU.jl), and Intel (OneAPI.jl)
### Feature Request

A similar approach to what Flux.jl does [here](https://github.com/FluxML/Flux.jl/tree/master/ext) could be used to extend GPU support to all backends supported by KernelAbstractions.jl. #179 looks like a good starting point to work from.

rkierulf:
This pull request adds support for GPU backends Metal, ROCm, and oneAPI with the corresponding Julia packages (Metal.jl, AMDGPU.jl, oneAPI.jl). CUDA is still supported but is no longer an explicit dependency, so users will now need to add 'using CUDA' to their code or they will see a message informing them that no GPU backends are loaded. I've run the KomaMRICore tests with CUDA and Metal and ensured that they pass, but for AMD and oneAPI we will need to wait until 
#147   to confirm that this works as expected.

I haven't made any documentation updates yet, so let me know if that needs to be done before merging. The last thing to note is with the ArbitraryMotion struct, there was a custom adapt_storage function for adapting to a CUDA device that I wasn't sure whether it was necessary, since I think Adapt.jl and Functors.jl should be able to adapt the fields correctly and the ArbitraryMotion GPU test still passed with my changes in place. Nevertheless, it's possible I will need to add that function back, and define adapt_storage functions for converting ArbitraryMotion to the other backends.


cncastillo (changes requested):
Perfect! add `print_devices` in `docs/src/reference/3-koma-core.md` so the documentation doesn't fail.

Maybe something like this (I renamed the subsection)

```md
## GPU helper functions

```@docs
print_devices
gpu
cpu
f32
f64```
```

Should we add a warning to `KomaMetalExt.__init__` saying that performance could be affected by https://github.com/JuliaGPU/Metal.jl/issues/348?

cncastillo (approved):
Amazing :smiley: !! well done!

PR #411, opened 2024-06-17
Add Buildkite GPU pipeline

rkierulf:
This is an initial implementation for creating a Buildkite pipeline to test on all 4 GPU backends (CUDA, AMD, Metal and oneAPI). As discussed with Carlos, I simplified the KomaMRICore tests so there are no longer separate tests for CPU single-thread, CPU multi-thread, and GPU. Instead, the tests now look to see if there is a test argument for "AMDGPU", "CUDA", "Metal", or "oneAPI", and if so load the corresponding package, which will trigger loading the corresponding ext module from KomaMRICore. The tests no longer set anything for the "Nthreads" parameter, so they can be made to run with a specific number of CPU threads by passing the julia argument --threads=<nthreads> to Pkg.test(). For now, the SimpleMotion and ArbitraryMotion tests still have the :skipci tag since there is at least one issue with SimpleMotion on Metal that has not yet been resolved, I still need to check to see if it has been fixed by #408 .


.buildkite/pipeline.yml
```diff
@@ -1,5 +1,117 @@
...
+          version: "{{matrix.version}}"
+    command:
+      julia -e 'println("--- :julia: Instantiating project")
+          using Pkg
+          Pkg.develop([
+              PackageSpec(path=pwd(), subdir="."),
+              PackageSpec(path=pwd(), subdir="KomaMRIBase"),
+              PackageSpec(path=pwd(), subdir="KomaMRICore"),
+              PackageSpec(path=pwd(), subdir="KomaMRIFiles"),
+              PackageSpec(path=pwd(), subdir="KomaMRIPlots")
```
cncastillo, line 20:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,117 @@
...
```
cncastillo, line 49:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,117 @@
...
```
cncastillo, line 78:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,117 @@
...
```
cncastillo, line 108:
```suggestion
```

KomaMRICore/test/runtests.jl
```diff
@@ -1,6 +1,27 @@
...
+#   import Pkg
+#   Pkg.test("KomaMRICore"; julia_args=`--threads=4`)
+###
+
+#Environment variable set by CI
+const CI = get(ENV, "CI", nothing)
+
+@run_package_tests filter=ti->(:core in ti.tags)&&(isnothing(CI) || :skipci ∉ ti.tags) #verbose=true
 
 @testitem "Spinors×Mag" tags=[:core] begin
```
cncastillo, line 26:
```suggestion
@testitem "Spinors×Mag" tags=[:core, :skipbuildkite] begin
```

KomaMRICore/test/runtests.jl
```diff
@@ -136,13 +157,15 @@ end
...
+    include("initialize.jl")
 
     seq = PulseDesigner.EPI_example()
     sys = Scanner()
     obj = brain_phantom2D()
 
     sim_params = KomaMRICore.default_sim_params()
     sim_params["return_type"] = "mat"
+    sim_params["gpu"] = USE_GPU
     sig = @suppress simulate(obj, seq, sys; sim_params)
```
cncastillo, line 169:
```suggestion
@testitem "signal_to_raw_data" tags=[:core, :skipbuildkite] begin
    using Suppressor

    seq = PulseDesigner.EPI_example()
    sys = Scanner()
    obj = brain_phantom2D()

    sim_params = KomaMRICore.default_sim_params()
    sim_params["return_type"] = "mat"
    sig = @suppress simulate(obj, seq, sys; sim_params)
```

KomaMRICore/test/runtests.jl
```diff
@@ -374,6 +257,7 @@ end
 
 @testitem "Bloch_phase_compensation" tags=[:important, :core] begin
     using Suppressor
+    include("initialize.jl")
```
cncastillo, line 260:
```suggestion
@testitem "Bloch_phase_compensation" tags=[:important, :core, :skipbuildkite] begin
    using Suppressor
```

KomaMRICore/test/runtests.jl
```diff
@@ -395,16 +279,16 @@ end
     seq2 += RF(B1 .* exp(1im*rf_phase), Trf)
     seq2 += ADC(N, Tadc, 0, 0, rf_phase)
 
-    sim_params = Dict{String, Any}("Δt_rf"=>1e-5, "gpu"=>false, "Nthreads"=>1)
+    sim_params = Dict{String, Any}("Δt_rf"=>1e-5, "gpu"=>USE_GPU)
```
cncastillo, line 282:
```suggestion
    sim_params = Dict{String, Any}("Δt_rf"=>1e-5)
```

KomaMRICore/test/runtests.jl
```diff
@@ -577,6 +362,7 @@ end
 
 @testitem "simulate_slice_profile" tags=[:core] begin
     using Suppressor
+    include("initialize.jl")
```
cncastillo, line 365:
```suggestion
@testitem "simulate_slice_profile" tags=[:core, :skipbuildkite] begin
    using Suppressor
```

KomaMRICore/test/runtests.jl
```diff
@@ -591,7 +377,9 @@ end
     seq = PulseDesigner.RF_sinc(B1, Trf, sys; G=[0; 0; Gz], TBP=8)
 
     # Simulate the slice profile
-    sim_params = Dict{String, Any}("Δt_rf" => Trf / length(seq.RF.A[1]))
+    sim_params = Dict{String, Any}(
+        "Δt_rf" => Trf / length(seq.RF.A[1]),
+        "gpu" => USE_GPU)
```
cncastillo, line 382:
```suggestion
    sim_params = Dict{String, Any}("Δt_rf" => Trf / length(seq.RF.A[1]))
```

KomaMRICore/test/runtests.jl
```diff
@@ -1,6 +1,27 @@
...
+#   Example:
+#   
+#   import Pkg
+#   Pkg.test("KomaMRICore"; julia_args=`--threads=4`)
+###
+
+#Environment variable set by CI
+const CI = get(ENV, "CI", nothing)
+
+@run_package_tests filter=ti->(:core in ti.tags)&&(isnothing(CI) || :skipci ∉ ti.tags) #verbose=true
```
cncastillo, line 24:
```suggestion
is_local_run = isnothing(CI) 
is_buildkite_run = CI == "BUILDKITE"

@run_package_tests filter=ti->(:core in ti.tags) && (is_local_run || :skipci ∉ ti.tags) && (!is_buildkite_run || :skipbuildkite ∉ ti.tags)
```

The logic might me wrong :sweat_smile:.

KomaMRICore/test/initialize.jl
cncastillo, line 1: Any way we can select which test to run in VSCode?

cncastillo (approved):
Amazing! I love the PRs that remove code :smile:, well done!!

 I added a few code suggestions, so we can filter with `:skipbuildkite`, as some tests are pretty GPU agnostic. Also, I think that "KomaMRIFiels" and "KomaMRIPlots" are not used for these tests, so I suggested removing them.

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,117 @@
...
+    matrix:
+      setup:
+        version:
+          - "1.9"
+    env:
+      CI: BUILDKITE
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "{{matrix.version}}"
+    command:
```
cncastillo, line 70:
```suggestion
    command: |
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,117 @@
...
```
cncastillo, line 100:
```suggestion
    command: |
```

cncastillo (changes requested):
I applied the code review changes, but some stuff from AMD needs Julia 1.10, so unfortunately I think we will need to use Julia 1.10 for that one. OneAPI seems to have the same problem with `cumsum` as Metal.

CUDA and Metal are working! https://buildkite.com/julialang/komamri-dot-jl/builds/593

I saw the CUDA ext being loaded in AMD, maybe we need to put them as weakdeps in the `tests/Project.toml`?

KomaMRICore/test/initialize.jl
rkierulf: I don't think so since there doesn't appear to be a way to pass test arguments through the VSCode test panel. If a user wants to run an individual test on a specific backend, the easiest way to do it might be to add this:

ARGS = ["CUDA"]

before the line with include("initialize.jl")

rkierulf:
I changed the Julia version for the AMD tests to 1.10 and added the same workaround we're using for Metal into oneAPIExt.jl. 

For the test Project.toml dependencies, I tried moving the GPU packages to [weakdeps] but it complained about them not being loaded when I tried to run the tests, even if I included them with 'using' before. DiffEqGPU.jl is adding the packages and then deleting test/Manifest.toml to avoid having any of the packages as explicit dependencies: https://github.com/SciML/DiffEqGPU.jl/blob/master/.buildkite/runtests.yml, so we could try and follow what they're doing, however, the problem I see with this is that I'm not sure how a user would be able to run tests on their own GPU without changing the test/Project.toml themselves, which may not be desirable.

It looks like AMD now has an issue with the device name function (should be easy to fix), and Intel has an issue with 'unsupported use of double value'.

rkierulf:
Nice, AMD is now passing!

The Intel error may take a bit more time to resolve, I'll look into tomorrow

cncastillo:
This PR could be relevant:
 - https://github.com/FluxML/Flux.jl/pull/2453

KomaMRICore/test/initialize.jl
cncastillo: Mmm maybe using Preferences.jl?

rkierulf: I think Preferences.jl requires being inside a module that corresponds to a loaded package, and with the way TestItems.jl works it creates separate modules for each TestItem. If I try to do anything with Preferences.jl inside run_tests.jl or initialize.jl I get an error: 'ArgumentError: Module Main does not correspond to a loaded package!'. There might be a way to get this to work, but I'm not sure how at the moment.

rkierulf:
I copied the approach from that pull request so we are no longer installing all GPU backends. I also added a new gpu method without any backend parameter that uses the logic in get_backend() to determine which backend to use. The oneAPI tests are still failing on this line in BlochSimulationMethod.jl:

Mxy  = [M.xy M.xy .* exp.(1im .* ϕ .- tp' ./ p.T2)]

I think there is a Float64 value sneaking in somehow based on the 'unsupported use of double value' message.

KomaMRICore/test/initialize.jl
cncastillo: Oh, it is fine then. If by default it tries to load CUDA, I think that would be enough, like `isempty(ARGS) && ARGS = ["CUDA"]`.

rkierulf:
oneAPI is passing now that I switched the line with:

exp.(exp_argument)

to the equivalent:

exp.(real.(exp_argument)) .* (cos.(imag.(exp_argument)) .+ Complex{T}(0,1) * sin.(imag.(exp_argument)))

I'm pretty sure this is a oneAPI bug, but nice that we can work around it for now.

cncastillo:
We will be able to use exp when this is merged :)
- https://github.com/JuliaGPU/oneAPI.jl/pull/443

cncastillo:
We also got a passing badge! 

[Build status](https://buildkite.com/julialang/komamri-dot-jl)

cncastillo:
FYI:
- https://github.com/JuliaGPU/Metal.jl/issues/348
- https://github.com/JuliaGPU/Metal.jl/pull/366

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,127 @@
...
+        version:
+          - "1.9"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "{{matrix.version}}"
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+          dirs:
+            - KomaMRICore/src
+            - KomaMRICore/ext
```
cncastillo, line 74:
```suggestion
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,127 @@
...
+          using Pkg
+          Pkg.develop([
+              PackageSpec(path=pwd(), subdir="."),
+              PackageSpec(path=pwd(), subdir="KomaMRIBase"),
+              PackageSpec(path=pwd(), subdir="KomaMRICore"),
+          ])'
+      
+      julia -e 'println("--- :julia: Running tests")
+          using Pkg
+          Pkg.test("KomaMRICore"; coverage=true, test_args=["Metal"])'
```
cncastillo, line 86:
```suggestion
          Pkg.test("KomaMRICore"; test_args=["Metal"])'
```

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,127 @@
...
+    agents:
+      queue: "juliagpu"
+      intel: "*"
+    timeout_in_minutes: 120
+
+env:
+  CI: BUILDKITE
+  CODECOV_FLAGS: core
+  JULIA_PKG_SERVER: ""
+  SECRET_CODECOV_TOKEN: "uBxOIgxCT1D0dg7NKagIb8tZgXjRYO1Rdvqf0O57c5Opt3+QzBp596B09KZbxsZXqdIfFwpAtjA2f+Aklhq2yNdkCoyolPkA+1zGrhOVgdXxvHc7xZUw+S0z8hBcftN50kYo+B5Vh5TmyxeXiPYNyLpGI3Y/2L9lRwdgIzzdQjzUsEy9HrkvdGuzgvUa3w46XP5yxkYIHihb9qPAL/VMEmqpvRURhOi2TO9GBoT3ussCe542OkC/FhlPKjWGIWFCpq0+DkJRgcBw6Oi+lvX0+bNmcGhHA0V8bLA3+++lCfyXTrSrbicq21uEcMO1S0cCPzLSt96V46Tc0bpb4UCcpQ==;U2FsdGVkX180D1vvwchhVGlm7wHnbtlB+sD+4UM5D2pUuaF5YEcbH6yUTsZ1EBGUxnMmQrMCM/o9BM28LitpMg=="
```
cncastillo, line 127:
```suggestion
  SECRET_CODECOV_TOKEN: "lUmUVYkTlE8u0mR/ymv5rtE1A59wXZmQ3miRkmKciC/4+xHPNjpeIJ03FupuC1ElGBeX0m6DDFavZ9burLosGxbBYIPziBQZ5P9NdPDZjBdo7NM3QSSBeUfDsDYbHsYglfJZ35UL6Pd2YTAkJG0ePrTpfUaBb9rcll926NdqUP0vE2hbR2leKFFgBVNtK9Zf+NE7hO3meZQEZ+sN5tA7xGr24A3Ay7ckPg5HbPPD3KII2/fLtW+w0fQUfJdXNFrajJ0FyNE0kNLDlIzoYN6XM4yUeLBXoyXFHnqT5dSu7pqrByEk/ptYpjUzKFoRSnDQy+p8vrx9e/iAVf3lwwgwWA==;U2FsdGVkX1+IZLGTj2FNEy4XvUQbzkZFEwUpwIdRDCKrEea5O/OKGV5vYJufty3m9yniE+av4937HabS9dO1RA=="
```

cncastillo (approved):
Due to problems with the codecov keys for Metal, we decided to remove Metal's codecov :disappointed: .

PR #413, opened 2024-06-18
Fix plot phantom function

### LINKED ISSUE #407: [KomaUI Plotting] obj_ui[] does not draw on first time updated. Need to press :rho or other option.
### What happened?

obj_ui[] does not draw on first time updated: 
obj_ui_first_update

Need to press :rho or other button option:
push_rho

### Environment

```Markdown
OS x86_64-linux-gnu
Julia 1.10.4
KomaMRIPlots 0.8.3
KomaMRIFiles 0.8.2
KomaMRI 0.8.2
KomaMRICore 0.8.3
KomaMRIBase 0.8.4
```

pvillacorta:
This PR is intended to fix #407 
I also correct some typos in the docstrings for some SimpleMotion types

### DIFF
--- a/KomaMRIBase/src/datatypes/phantom/motion/simplemotion/HeartBeat.jl
+++ b/KomaMRIBase/src/datatypes/phantom/motion/simplemotion/HeartBeat.jl
@@ -1,5 +1,5 @@
 @doc raw"""
-    heartbeat = HeartBeat(t_start, t_end, dx, dy, dz)
+    heartbeat = HeartBeat(circumferential_strain, radial_strain, longitudinal_strain, t_start, t_end)
 
 HeartBeat struct. It produces a heartbeat-like motion, characterised by three types of strain:
 Circumferential, Radial and Longitudinal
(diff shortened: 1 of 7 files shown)


cncastillo:
Hi I don't think this solves the problem in #407 yet. I think you need to think about how to use `plot_koma` properly and redefine it if you need to. The plots should work with no problems for `NoMotion` and `SimpleMotion/ArbitraryMotion` for:

- [ ] Julia REPL
- [ ] VScode
- [ ] Pluto
- [ ] Jupyter
- [ ] KomaUI

My understanding is that now, `plot_phantom_map` with motion has problems with some of those. 

The main problem is that **the current implementation also breaks plots without motion**. For phantoms with `NoMotion,` the function should do exactly what was done before (like for the simulation), just a few minor modifications (like the spin subsampling for example). In the meantime, I would only use `Plot` for the motion plots, and `plot_koma` for the others (dispatch), sharing some code for the init as functions (no function definitions inside another function).

- [ ] use `Plot` for the motion plots, and `plot_koma` for the others (dispatch `plot_phantom_map` based on `Phantom{MotionModel}`)
- [ ] sharing some code for the init as functions (no function definitions inside another function)

PR #416, opened 2024-06-21
Add kernel-based matrix cumsum

rkierulf:
I added a kernel implementation of cumsum(A::Matrix, dims=2). Since Metal and oneAPI don't yet support cumsum and findall, we are working around the issue for now by copying to the CPU, doing the operations there and then copying back. It looks like the 2D cumsum in the cumtrapz function of KomaMRIBase is the main place where this could be significantly slower than an actual GPU implementation (the other cumsum and findall operations are on 1D arrays), so I wrote a simple kernel equivalent to avoid having to copy to the CPU.

### DIFF
--- a/KomaMRICore/src/simulation/Bloch/KernelFunctions.jl
+++ b/KomaMRICore/src/simulation/Bloch/KernelFunctions.jl
@@ -1,6 +1,7 @@
 using KernelAbstractions: @index, @kernel
 
 ## COV_EXCL_START
+#! format: off
 
 """
     cumsum2_kernel
@@ -19,6 +20,7 @@ in-place on the input matrix without allocating additional memory.
     end
 end
 
+#! format: on
 ## COV_EXCL_STOP
 
 """
--- a/KomaMRICore/src/simulation/Bloch/KernelFunctions.jl
+++ b/KomaMRICore/src/simulation/Bloch/KernelFunctions.jl
@@ -1,5 +1,7 @@
 using KernelAbstractions: @index, @kernel
 
+## COV_EXCL_START
+
 """
     cumsum2_kernel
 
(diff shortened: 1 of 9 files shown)


KomaMRICore/src/simulation/Bloch/KernelFunctions.jl
```diff
@@ -0,0 +1,48 @@
+using KernelAbstractions: @index, @kernel
```
cncastillo, line 1: Can you check if the newly added CPU backend adds codecov for the kernel?

Also, add the comments to exclude the coverage for the metal extension.

You may also delete the interpolation adapters from the GPU extensions.

rkierulf:
I added some more tests to increase code coverage. Although it is still not reaching the target, a lot of the remaining uncovered lines are for custom GPU logic for the MotionModel structs which is being eliminated in #408 . Once this is done, I think it should reach the goal.

KomaMRICore/test/runtests.jl
```diff
@@ -385,3 +386,20 @@ end
...
+    import KernelAbstractions as KA
+    include("initialize.jl")
+
+    x = ones(Float32, 1000)
+    if USE_GPU
+        x = x |> gpu
+        @test KA.get_backend(x) isa KA.GPU
+    else
+        @test true
+    end
```
cncastillo, line 401: I think it would be good to check that

Use_gpu = true
- cpu(gpu(x)) is a CPU array
Use_gpu = false
- cpu and gpu are no-ops

cncastillo (approved):
I added a few comments that could increase the codecov further If it's not enough.

KomaMRICore/src/simulation/Bloch/KernelFunctions.jl
rkierulf: I don't think it does since on the CPU it forwards the call to cumtrapz in KomaMRIBase. I forgot about code coverage not working for Metal, so I added the comments to exclude.

PR #418, opened 2024-06-26
Initial benchmarking implementation: steps to run benchmarks and comb…

rkierulf:
This pull request creates a pipeline to run benchmarks on all supported backends, combine the results into a single json file, and pass that file to https://github.com/benchmark-action/github-action-benchmark. So far, I've written the benchmarking scripts and buildkite pipeline for running and combining the benchmarks. Once I confirm this is working, then I'll add the github step for downloading the resulting artifact and passing to github-action-benchmark. 

Currently, I have it running the CPU benchmarks with 1, 2, 4, and 8 threads, but we can change this if we want. The two benchmarks are the examples from lit-04-3DSliceSelective.jl and MRiLab_speed.jl, with names "Slice Selection 3D" and "MRI Lab" (we can change these names). I also noticed that the lit-04-3DSliceSelective.jl example takes kind of long to run (around a minute on my computer), which may cause BenchmarkTools.jl to run fewer trials and result in more fluctuation in our measurements, so we may want to adjust this example to not take quite as long.

We should also start thinking about how we want the benchmark results to display. Currently, I have the benchmark structure as:

benchmark -> sim method -> cpu / gpu -> # of threads / gpu backend

As of now, the only sim method is Bloch (is it interesting to also profile BlochDict?). When a kernel-based sim method is written we can add this as well. If we want our page to look similar to the one here: https://lux.csail.mit.edu/benchmarks/, we can have two columns for CPU and GPU, and the graphs will have a title like: "MRI Lab + Bloch + GPU", with lines for each backend (and # of threads for the CPU). This will require some updates to the index.html page that github-action-benchmark provides, but this can be done after this pull request is merged.


bench/aggregate.jl
```diff
@@ -0,0 +1,37 @@
...
+const oneAPI_results = BenchmarkTools.load(oneAPI_filepath)[1]
+
+@assert CPU_results isa BenchmarkTools.BenchmarkGroup
+@assert AMDGPU_results isa BenchmarkTools.BenchmarkGroup
+@assert CUDA_results isa BenchmarkTools.BenchmarkGroup
+@assert Metal_results isa BenchmarkTools.BenchmarkGroup
+@assert oneAPI_results isa BenchmarkTools.BenchmarkGroup
+
+# Add other results to CPU results
+for benchark in keys(CPU_results)
```
cncastillo, line 28:
```suggestion
for benchmark in keys(CPU_results)
```

cncastillo:
For convenience I will leave this link here: https://juliahealth.org/KomaMRI.jl/dev/bench/

rkierulf:
I think we can merge this tomorrow after I delete the lines with #TO-DO: remove this. I still want to see if there is a way we can run the CPU benchmarks on the same platform each time. 

We'll also need to ensure we delete the data.js file in gh-pages/benchmarks before merging.

PR #421, opened 2024-07-03
Updates to github-action-benchmark index.html

rkierulf:
I updated the default index.html provided by github-action-benchmark to more closely match Lux.jl's benchmark display. I also added an option at the top to view memory use and allocation, in addition to time, and deleted the /bench folder which was used to store the results for one commit before we changed the folder name to 'benchmark'. 

If you want to test these changes, I've been using the [VSCode live server extension](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer): just make sure the data.js and index.html pages are in the same directory, and with the index.html file open, right click -> Open with Live Server. Feel free to change the colors or any other display details.


cncastillo:
Nah I trust you :). Funny that the `buildkite/komamri-dot-jl` got triggered by this. I will be merging it.

PR #422, opened 2024-07-03
Fixing issues in Julia 1.11 and Julia 1.12

cncastillo:
The incorrect definition of the `getproperty` function for `Array{Grad/RF/ADC}` generated some weird errors. This PR fixes these issues.

---

I also included `[workspace]`'s to the Project.toml's. A feature that will be introduced in Julia 1.12 https://github.com/JuliaLang/Pkg.jl/pull/3841. The cool thing is that our mono repo and test setups are greatly simplified. For the KomaMRICore tests, the `test_args` are just `args`:

```julia
❯ julia --project=test --threads=4 test/runtests.jl
[ Info: Testing on the CPU with 4 thread(s)
...
```
and for CUDA (with CUDA added to the `v1.12` environment)
```julia
❯ julia --project=test test/runtests.jl CUDA
```
~~This one does not work yet, due to https://github.com/JuliaGPU/GPUCompiler.jl/issues/593.~~

---

The mono repo setup should be (I think?)

```julia
pkg> dev KomaMRI
```
Instead of dev'ing all the environments for each sub-package and their corresponding test folders.


cncastillo:
Apparently, the pre-release CI  is failing due to problems in MRIFiles. Check: https://github.com/MagneticResonanceImaging/MRIReco.jl/issues/194.

cncastillo:
I am going to merge this and look into MRIReco's problem later.

PR #431, opened 2024-07-09
Including details about prints of simulate

gsahonero:


cncastillo:
@rkierulf There seems to be a few minor issues with the benchmarks/tests.

Are we still running the codecov for Metal? (https://buildkite.com/julialang/komamri-dot-jl/builds/810)  

If the PR is from a fork and not coming from "us" the benchmark.yml (GitHub) job fails due to not having the `BUILDKITE_TOKEN`. Maybe we can suspend the benchmarks for external PRs. The benchmarks in Buildkite are not run.

rkierulf:
We shouldn't be running codecov for Metal since the step in runtests.yml is
missing the plugin. For the Benchmark.yml file, I think we can prevent the
action from running for external pull requests if we change the if
statement in line 35 to:

if: "!contains(github.event.head_commit.message, '[skip benchmarks]') && {{
github.event.pull_request.head.repo.full_name ==
github.event.pull_request.base.repo.full_name }}"

https://github.com/orgs/community/discussions/26829

For the codecov issue with the GPU tests, this seems to have started
recently, since none of the builds prior to #806 had this issue, but all of
the builds starting from yesterday (#807+) do. I don't see codecov
decryption errors for the other Julialang pipelines however, so it doesn't
seem to be a general Buildkite issue. Is there anything with the repository
that changed between Friday and yesterday?

Ryan

cncastillo:
I don't think I have changed anything :(. I think it could be related to a problem with the codecov_token in external PRs. Maybe something that I need to change in the Buildkite config (?).

KomaMRICore/src/simulation/SimulatorCore.jl
```diff
@@ -292,7 +292,13 @@ end
...
 
 Returns the raw signal or the last state of the magnetization according to the value
-of the `"return_type"` key of the `sim_params` dictionary.
+of the `"return_type"` key of the `sim_params` dictionary. By default, it prints the `@time`
+information of `run_sim_time_iter` and simulation details such as the number of spins,
+the simulation method, the time points to consider, and the ADC points to be acquired.
+
+Note: The information obtained through `@time out = simulate(obj, seq, sys)` will differ from 
+the report of `out = simulate(obj, seq, sys)` as the latter one does not consider complementary 
+processes to `run_sim_time_iter`.
```
cncastillo, line 301:
```suggestion
of the `"return_type"` key of the `sim_params` dictionary. 

This is a wrapper function to `run_sim_time_iter`, which converts the inputs to the appropriate types and discretizes the sequence before simulation. The reported simulation time only considers `run_sim_time_iter`, as the preprocessing duration should be negligible compared to the simulation time (if this is not the case, please file a bug report). 
```

cncastillo (approved):

Issue #410, opened 2024-06-14
New motion approach to combine SimpleMotion and ArbitraryMotion into the same phantom


pvillacorta:
Idea for new motion related types and structs redistribution, in order to be able to combine them and make them more alike:

``` julia
abstract type Motion end

abstract type SimpleMotion <: Motion end
abstract type ArbitraryMotion <: Motion end
```

Specific motion structures:
``` julia
Translation <: SimpleMotion
Rotation <: SimpleMotion
...
PeriodicHeartBeat <: SimpleMotion
```

``` julia
Trajectory <: ArbitraryMotion
PeriodicTrajectory <: ArbitraryMotion
FlowTrajectory <: ArbitraryMotion
PeriodicFlowTrajectory <: ArbitraryMotion
```

That way, phantom struct will look like this:
``` julia
@with_kw mutable struct Phantom{T<:Real}
    ...
    motion::Vector{<:Motion{T}} = []
    ...
end
```

and motion will be initialized, for example, like this:
``` julia
motion = [Translation(...), Rotation(...), Trajectory(...), HeartBeat(...)]
```

cncastillo:
This is a very good idea! We just need to think about how this will impact:
 - `obj1 + obj2`: `obj1` has `motion1` and `obj2` has `motion2` (one ball going up, and another one going down) 
 - `obj.phantom`: phantom file read/write

pvillacorta:
Another possible approach regarding periodicity:
```julia
abstract type Periodicity end

struct Periodic <: Periodicity end
struct Aperiodic <: Periodicity end

struct Translation{T<:Real, P<:Periodicity} <: SimpleMotionType{T}
   ...
end
```

pvillacorta:
Regarding flow:
```julia
function reset_magnetization!(M::Mag{T}, Mxy::AbstractArray{T}, motion::MotionModel{T})
   return nothing
end

function reset_magnetization!(M::Mag{T}, Mxy::AbstractArray{T}, motion::FlowTrajectory{T})
   # do things with flow (create interpolation functions, resolve them, etc):
   Mxy .= ...
   M.z .= ...
   return nothing
end
```

Issue #376, opened 2024-04-18
Define `Nspins` for SimpleMotion


pvillacorta:
It would be nice if two spin groups could have different Simple Motion's, for that SimpleMotion should have a `Nspins` that affects, so we can index it properly.

_Originally posted by @cncastillo in https://github.com/JuliaHealth/KomaMRI.jl/pull/184#discussion_r1562571617_

Issue #395, opened 2024-05-30
Add Phantom Explanation section


pvillacorta:
Explanations: Add "Phantom" section, with the following subsections:

- Phantom structure
- Motion: by the moment, SimpleMotion and simple motion types
- .phantom file format

pvillacorta:
> `motion <: AbstractMotionSet` I thought we changed `AbstractMotionSet` to `AbstractMotion`. Set has a precise mathematical meaning.
> 
> `AbstractMotion`:
> - `struct NoMotion`
> - `struct MotionList` (The same as an AbstractArray{Motion} but we didn't use arrays, because ... ? we will need to discuss this in the paper)
> 	- `struct Motion` (This one makes sense)
> 		- `Action <: AbstractAction`
> 		- `TimeSpan <: AbstractTimeSpan`
> 		- `SpinSpan <: AbstractSpinSpan` (Is the user meant to extend this one?)
> 		
> Is it like this? 
> 
> If I don't want to define a list, can I just define `obj.motion = Translate(...)`?
> 
> What function must I define to generate a new `AbstractMotion`/`AbstractAction`/...?
> 
> If I want to define my new custom motion, should create a new AbstractAction or a new AbstractMotion? 
> 
> Is ECG triggering possible? #73 
> 


Commented by @cncastillo

PR #442, opened 2024-07-11
New Motion

pvillacorta:
This PR is intended to address #410

Still to be done:
- [x] Revise and change docstrings
- [x] Improve tests
- [ ] Benchmarking (simulation times, allocations, accuracy...)
- [x] Test with flow phantoms
- [x] Remake phantoms from /examples/2.phantoms:
     - brain_motion
     - contracting_ring
     - flow phantom
- [x] Define the `Nspins` parameter for all motion types (I don't know if for this PR) -> EDIT: Now called `SpinRange` (solves [#376](https://github.com/JuliaHealth/KomaMRI.jl/issues/376))


rkierulf:
I think I've found the cause of the Simple and Arbitrary Motion test failures for Metal: in UnitTime.jl, this line is processed incorrectly:

``` return min.(max.((t .- t_start) ./ (t_end - t_start), zero(T)), oneunit(T)) ```

It seems as if Metal has issues compiling the min and max broadcast together. However, if they are separated then it works as expected:

``` 
tmp = max.((t .- t_start) ./ (t_end - t_start), zero(T))
return min.(tmp, oneunit(T))
```

Making this change fixed the failing tests on my computer. I've also written https://github.com/JuliaGPU/Metal.jl/issues/389 since this appears to be a Metal issue with Julia 1.10.

**Comment by cncastillo** [comment]
You can preview the docs here (when they run):
https://juliahealth.org/KomaMRI.jl/previews/PR442

**Review by cncastillo — CHANGES_REQUESTED** [review]
I left some minor comments, most of them are renaming stuff. The documentation part can be a different PR, but we may want to create an issue so we don't forget. At least let's make the CI happy.

Also, I'm not sure if writing/reading phantoms work. I assume it does as it is passing the tests.

Did you try Bloch and BlochSimple?

We may want to add a benchmark that includes motion.

File: KomaMRIBase/test/runtests.jl [hunk 1]
```diff
@@ -538,24 +538,24 @@ end
...
         @test yt == ph.y .+ dy
         @test zt == ph.z .+ dz
     end
 
-    simplemotion = SimpleMotion(
-        PeriodicTranslation(dx=0.05, dy=0.05, dz=0.0, period=0.5, asymmetry=0.5),
-        Rotation(pitch=0.0, roll=0.0, yaw=π / 2, t_start=0.05, t_end=0.5),
+    simplemotion = MotionVector(
+        Translation(times=Periodic(period=0.5, asymmetry=0.5), dx=0.05, dy=0.05, dz=0.0),
+        Rotation(times=TimeRange(t_start=0.05, t_end=0.5), pitch=0.0, roll=0.0, yaw=π / 2)
```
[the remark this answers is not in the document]
pvillacorta: > Motion to AbstractMotion (what methods need to be defined to create a new one? Docs)

I cannot do this since AbstractMotion is already defined as an abstract type:
```julia
struct MotionVector{T<:Real} <: AbstractMotion{T}
    motions::Vector{<:Motion{T}}
end
```
```julia
struct NoMotion{T<:Real} <: AbstractMotion{T} end
```
I will explain you why I needed to define this abstract type.
Ok with the rest of suggested changes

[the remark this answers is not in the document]
cncastillo: There are two options

(1) rename the current `AbstractMotion` to something else, as it is not meant to be extended and the name makes more sense for `Translation`, `Rotation`, `Flow`, etc.

(2) remove the reason for its existence, like making `NoMotion{T<:Real}` an alias for an empty `MotionList` so we can dispatch based on it.

but I could be missing something.

cncastillo:
Btw, the min(max) problem was already fixed in GPUCompiler.jl, thanks to Ryan's report. So this is no longer necessary:

```julia
tmp = max.((t .- t_start) ./ (t_end - t_start), zero(T))
return min.(tmp, oneunit(T))
```

pvillacorta:
> Btw, the min(max) problem was already fixed in GPUCompiler.jl ([JuliaGPU/GPUCompiler.jl#602](https://github.com/JuliaGPU/GPUCompiler.jl/pull/602)), thanks to Ryan's report. So this is no longer necessary:
> 
> ```julia
> tmp = max.((t .- t_start) ./ (t_end - t_start), zero(T))
> return min.(tmp, oneunit(T))
> ```

How can I apply those changes? Tests are still failing if I don't define the `tmp` variable

cncastillo:
Nothing, the latest version of all the packages will be used.

In the CI there are some errors due to renaming 

```julia
LoadError: UndefVarError: `AbstractMotionList` not defined
```

cncastillo:
Metal and oneAPI changed their compact to Julia 1.10, so we may want to do the same.

KomaMRIBase/src/datatypes/phantom/motion/arbitrarymotion/FlowTrajectory.jl
```diff
@@ -0,0 +1,7 @@
+struct FlowTrajectory{T<:Real, TS<:TimeScale{T}} <: ArbitraryMotion{T}
+    times::TS
+    dx::AbstractArray{T}
+    dy::AbstractArray{T}
+    dz::AbstractArray{T}
+    resetmag::AbstractArray{Bool}
```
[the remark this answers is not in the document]
pvillacorta: Renamed to `spin_reset`

KomaMRIBase/test/runtests.jl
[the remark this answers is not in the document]
pvillacorta: I have opted for the first option:
```julia
abstract type AbstractMotionList{T<:Real} end
```

Motion:
```julia
abstract type AbstractMotion{T<:Real} end

struct MotionList{T<:Real} <: AbstractMotionList{T}
    motions::Vector{<:AbstractMotion{T}}
end
```

NoMotion:
```julia
struct NoMotion{T<:Real} <: AbstractMotionList{T} end
```

pvillacorta:
oneAPI is now the one that fails. I should revise it

cncastillo:
Well, we can only merge it if it is passing CI. 

I only see it pass once; where are all the tests running there?

If it is a bug with oneAPI we need to find it and report it. That way we can make it better for everyone.

cncastillo:
I see that the process is getting killed for the last test. 

https://buildkite.com/julialang/komamri-dot-jl/builds/943#0191709c-718a-4cce-b72f-066f174fa3a5

I remember that something similar happened to @rkierulf for the benchmarks, and he made the test smaller (less spins) and the problem disappeared, not sure if it is related. Maybe the tests for one API are being timed out.

cncastillo:
I re-run the failing jobs and it seems to be working, not sure why the accuracy problem disappeared, might me related to a package updating. 

Can you check if this is the case? and fix the last requested change? 

Also, it might be worth separating the motion tests (filter :core and :motion) like they do in Oceanigans.jl. So, the core and motion tests run in parallel, and it is more apparent if the CI fails because of motion-related problems.

https://buildkite.com/clima/oceananigans/builds/17089#01916b50-f61d-4509-ac44-0efa0f5ea846

Or at least reduce the number of motion tests, one for Trajectory and another for Translation, compared to JEMRIS seems good. I believe that now I see four.

pvillacorta:
Commit https://github.com/JuliaHealth/KomaMRI.jl/pull/442/commits/29950749514a80af07a6bfa5859eb2e1059a014b makes a "big" change in order to include spin ranges for motions. A little redistribution of structures, files and directories has been made. 

Fixing the last requested change is pending yet, as well as documentation, one or two more days... :)

pvillacorta:
Oh, now it looks like it's failing for CPU haha, I forgot to test it on my PC before commiting. I will solve it soon

KomaMRIBase/src/datatypes/phantom/Motion.jl
```diff
@@ -0,0 +1,120 @@
...
+function Base.:(==)(mv1::MotionVector{T}, mv2::MotionVector{T}) where {T<:Real}
+    sort_motions!(mv1)
+    sort_motions!(mv2)
+    return reduce(&, mv1.motions .== mv2.motions)
+end
+function Base.:(≈)(mv1::MotionVector{T}, mv2::MotionVector{T}) where {T<:Real} 
+    sort_motions!(mv1)
+    sort_motions!(mv2)
+    return reduce(&, mv1.motions .≈ mv2.motions)
+end
```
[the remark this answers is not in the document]
pvillacorta: Solved in [c63aa40](https://github.com/JuliaHealth/KomaMRI.jl/pull/442/commits/c63aa40e13cbdb1357d2729ad2f118d55e2dbef8)

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -0,0 +1,29 @@
...
+      t_unit = KomaMRIBase.unit_time(t, m.time)
+      idx = KomaMRIBase.get_idx(m.spins)
+      reset_magnetization!(@view(M[idx]), @view(Mxy[idx, :]), m.action, t_unit, @view(ρ[idx]))
+   end
+   return nothing
+end
+
+function reset_magnetization!(M::Mag{T}, Mxy::AbstractArray{Complex{T}}, action::KomaMRIBase.AbstractActionSpan{T}, t, ρ) where {T<:Real}
+   return nothing
+end
```
cncastillo, line 19: I don't think you need to create two methods that `return nothing`. Just remove the type in `action/motion`. Is `t::AbstractArray{T}` needed?

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -0,0 +1,29 @@
...
+
+function reset_magnetization!(M::Mag{T}, Mxy::AbstractArray{Complex{T}}, action::FlowPath{T}, t, ρ) where {T<:Real}
+   itp = KomaMRIBase.interpolate(action.spin_reset, KomaMRIBase.Gridded(KomaMRIBase.Constant{KomaMRIBase.Previous}()), Val(size(action.spin_reset, 1)))
+   flags = KomaMRIBase.resample(itp, t)
+   reset = vec(any(flags .> 0; dims=2))
+   flags = (cumsum(flags; dims=2) .== 0)
+   Mxy .*= flags
+   M.z[reset] = ρ[reset]
+   return nothing
+end
```
cncastillo, line 29: Does this work with the new changes made by Ryan? The use of `Mxy` instead of `M.xy` seems strange.

KomaMRICore/src/simulation/Functors.jl
```diff
@@ -99,19 +97,23 @@ See also [`f32`](@ref).
 """
 f64(m) = paramtype(Float64, m)
 
+# Koma motion-related adapts
+adapt_storage(backend::KA.GPU, xs::MotionList) = MotionList(gpu.(xs.motions, Ref(backend)))
+adapt_storage(backend::KA.GPU, xs::Motion) = Motion(gpu(xs.action, backend), gpu(xs.time, backend), xs.spins)
```
cncastillo, line 102: I don't think calling `gpu` is correct here, use adapt, no?

KomaMRICore/src/simulation/SimMethods/Bloch/BlochGPU.jl
```diff
@@ -190,7 +189,7 @@ function run_spin_excitation!(
     pre.ΔT2 .= exp.(-seq.Δt' ./ p.T2)
 
     #Excitation
-    apply_excitation!(backend, 512)(M.xy, M.z, pre.φ, seq.B1, pre.Bz, pre.B, pre.ΔT1, pre.ΔT2, p.ρ, ndrange=size(M.xy,1))
+    apply_excitation!(backend, 256)(M.xy, M.z, pre.φ, seq.B1, pre.Bz, pre.B, pre.ΔT1, pre.ΔT2, p.ρ, ndrange=size(M.xy,1))
```
cncastillo, line 192: Launch config.

KomaMRICore/src/simulation/SimMethods/Magnetization.jl
```diff
@@ -21,6 +21,9 @@ Base.getindex(M::Mag, i::Integer) = Mag(M.xy[i,:], M.z[i,:])
 # M[a:b]
 Base.getindex(M::Mag, i::UnitRange) = Mag(M.xy[i], M.z[i])
 Base.view(M::Mag, i::UnitRange) = @views Mag(M.xy[i], M.z[i])
+# M[:]
+Base.getindex(M::Mag, i::Colon) = M[1:length(M.z)]
+Base.view(M::Mag, i::Colon) = @view(M[1:length(M.z)])
```
cncastillo, line 26: Would removing the UnitRange from the first definition help to not create more?

KomaMRICore/src/simulation/SimMethods/SimulationMethod.jl
```diff
@@ -21,6 +21,7 @@ function initialize_spins_state(
     Mxy = zeros(T, Nspins)
     Mz = obj.ρ
     Xt = Mag{T}(Mxy, Mz)
+    sort_motions!(obj.motion)
```
cncastillo, line 24: I think this should be in get_spin_coords for the reasons named before.

KomaMRICore/test/runtests.jl
```diff
@@ -458,3 +419,89 @@ end
...
+    ]
+
+    x0 = [0.1]
+    y0 = [0.1]
+    z0 = [0.0]
+
+    for m in motions
+        motion = MotionList(m)
+
+        coords(t) = get_spin_coords(motion, x0, y0, z0, t)
```
cncastillo, line 459: `x(t), y(t), z(t) = get_spin_coords(motion, x0, y0, z0, t)` does this work?

KomaMRICore/test/runtests.jl
```diff
@@ -458,3 +419,89 @@ end
...
+        # Scanner
+        sys = Scanner()
+        # Simulation
+        for sim_method in [KomaMRICore.Bloch(), KomaMRICore.BlochSimple(), KomaMRICore.BlochDict()]
+            sim_params = Dict{String, Any}(
+                "sim_method"=>sim_method,
+                "return_type"=>"mat"
+            )
+            raw_aux = @suppress simulate(obj, seq, sys; sim_params)
+            raw = raw_aux[:, 1, 1]
```
cncastillo, line 501: Add GPU variable in sim_params.
```suggestion
        for sim_method in [KomaMRICore.Bloch(), KomaMRICore.BlochSimple(), KomaMRICore.BlochDict()]
            sim_params = Dict{String, Any}(
                "sim_method"=>sim_method,
                "return_type"=>"mat"
            )
            raw = @suppress simulate(obj, seq, sys; sim_params)[:, 1, 1]
```
Also, it would not be considered if I create a new simulation method.

KomaMRIFiles/src/Phantom/Phantom.jl
```diff
@@ -5,108 +5,109 @@ Reads a (.phantom) file and creates a Phantom structure from it
...
     dims = read_attribute(fid, "Dims")
     Ns = read_attribute(fid, "Ns")
+    # Version
+    file_version = VersionNumber(read_attribute(fid, "Version"))
+    program_version = pkgversion(KomaMRIFiles)
+    if file_version.major != program_version.major | file_version.minor != program_version.minor
+        @warn "Version mismatch detected:
+         File version: $file_version
+         KomaMRIFiles version: $program_version
+         This may lead to compatibility issues. Please update the file or the program to the matching version."
```
cncastillo, line 17: This message could be shorter. And not use new lines. Also, you probably want to check the major version.

KomaMRIFiles/src/Phantom/Phantom.jl
```diff
@@ -5,108 +5,109 @@ Reads a (.phantom) file and creates a Phantom structure from it
...
 """
 function write_phantom(
-    # By the moment, only "Explicit" type 
-    # is considered when writing .phantom files
     obj::Phantom,
     filename::String;
     store_coords=[:x, :y, :z],
-    store_contrasts=[:ρ, :T1, :T2, :T2s, :Δw],
-    store_motion=true,
+    store_contrasts=[:ρ, :T1, :T2, :T2s, :Δw]
```
cncastillo, line 110: What if I don't want to store the motion?

KomaMRICore/test/runtests.jl
```diff
@@ -458,3 +419,89 @@ end
...
+        seq.GR[1,1] = Grad(Gx, duration)
+        seq.GR[2,1] = Grad(Gy, duration)
+        seq.GR[3,1] = Grad(Gz, duration)
+        seq.ADC[1] = ADC(Nadc, duration-Trf, Trf)
+        # Creating object
+        obj = Phantom(x = x0, y = y0, z = z0, ρ = [M0], T1 = [T1], T2 = [T2], motion = motion)
+        # Scanner
+        sys = Scanner()
+        # Simulation
+        for sim_method in [KomaMRICore.Bloch(), KomaMRICore.BlochSimple(), KomaMRICore.BlochDict()]
```
cncastillo, line 495: Follow the same structure as in nomotion, one test item per method, add testsets for arbitrary and simple motions. So we get more information if something fails.

cncastillo (changes requested):

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -54,9 +47,12 @@ julia> obj.ρ
     Dθ::AbstractVector{T}  = zeros(eltype(x), size(x))
     #Diff::Vector{DiffusionModel}  #Diffusion map
     #Motion
-    motion::MotionModel{T} = NoMotion{eltype(x)}()
+    motion::AbstractMotionSet{T} = NoMotion{eltype(x)}() 
 end
 
+non_string_phantom_fields = Iterators.filter(x -> fieldtype(Phantom, x) != String,         fieldnames(Phantom))
+vector_phantom_fields     = Iterators.filter(x -> fieldtype(Phantom, x) <: AbstractVector, fieldnames(Phantom))
```
cncastillo, line 54: Make constant `const` and put name in caps `VECTOR_PHANTOM_FIELDS`

KomaMRIBase/src/motion/motionlist/SpinSpan.jl
```diff
@@ -0,0 +1,71 @@
...
+```julia-repl
+julia> allspins = AllSpins()
+```
+"""
+struct AllSpins <: AbstractSpinSpan end
+
+# Functions
+Base.getindex(spins::AllSpins, p) = p, spins
+Base.view(spins::AllSpins, p) = p, spins
+get_idx(spins::AllSpins) = Colon()
```
cncastillo, line 22: Is this used?

KomaMRIBase/src/motion/motionlist/SpinSpan.jl
```diff
@@ -0,0 +1,71 @@
...
+    idx = intersect_idx(spins.range, p)
+    return intersect_idx(p, spins.range), SpinRange(idx)
+end
+function Base.view(spins::SpinRange, p)
+    idx = intersect_idx(spins.range, p)
+    return intersect_idx(p, spins.range), SpinRange(idx)
+end
+Base.:(==)(sr1::SpinRange, sr2::SpinRange) = sr1.range == sr2.range
+Base.length(sr::SpinRange) = length(sr.range)
+get_idx(spins::SpinRange) = spins.range
```
cncastillo, line 69: Same for this one

KomaMRICore/Project.toml
```diff
@@ -8,6 +8,7 @@ Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 KomaMRIBase = "d0bc0b20-b151-4d03-b2a4-6ca51751cb9c"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
```
cncastillo, line 11: This is still here?

KomaMRICore/src/simulation/Functors.jl
```diff
@@ -53,7 +52,7 @@ x = gpu(x, CUDABackend())
 ```
 """
 function gpu(x, backend::KA.GPU)
-    return fmap(x -> adapt(backend, x), x; exclude=_isleaf)
+    fmap(x -> adapt(backend, x), x; exclude=_isleaf)
```
cncastillo, line 55: revert, the `return` is needed by Julia blue guidelines.

KomaMRIFiles/src/Phantom/Phantom.jl
```diff
@@ -5,108 +5,108 @@ Reads a (.phantom) file and creates a Phantom structure from it
...
-    version = read_attribute(fid, "Version")
     dims = read_attribute(fid, "Dims")
     Ns = read_attribute(fid, "Ns")
+    # Version
+    file_version = VersionNumber(read_attribute(fid, "Version"))
+    program_version = pkgversion(KomaMRIFiles)
+    if file_version.major != program_version.major
+        @warn "Version mismatch detected: $file_version (file) vs $program_version (Koma)
+         This may lead to compatibility issues. Please update the file or the program."
+    end
```
cncastillo, line 16:
```suggestion
    if file_version.major != program_version.major
        @warn "KomaMRIFiles: Version mismatch detected: $file_version (used to write .phantom) vs $program_version (installed)
         This may lead to compatibility issues. "
    end
```

cncastillo (changes requested):
Very nice! some minor additional comments.

cncastillo:
Also, I just remembered, but there needs to be a pretty important change in how the Phantoms are plotted. Please dispatch based on `Phantom{Motion}` or something similar to change `plot_phantom_map,` as we need to use `plot_koma` instead of `Plot,` as the last one breaks compatibility with VSCode and Pluto.

pvillacorta:
> Also @pvillacorta, I just remembered, but there needs to be a pretty important change in how the Phantoms are plotted. Please dispatch based on `Phantom{Motion}` or something similar to change `plot_phantom_map,` as we need to use `plot_koma` instead of `Plot,` as the last one breaks compatibility with VSCode and Pluto. This is more important than all the other comments.

@cncastillo Okay, I will take this into account. I remember when I tried to solve this problem with the `Plot` and `plot_koma`: the main problem was the time slider, which somehow did not work with the `plot` function (the one called by `koma_plot`), and only worked with `Plot`. So there is two options for now:
1. Remove the slider. The plot will still be able to move, but it would not be possible to control the time.
2. Dispatch. Use `plot_koma` for phantoms with NoMotion, and `Plot` for dynamic phantoms.

In terms of code stability, option 1 is better, with the disadvantage that we would lose the slider...
Tell me what you think

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -0,0 +1,44 @@
...
+   t1 = @view(seq_t[1])
+   return add_t0 ? [t1 (t1 .+ t)] : t1 .+ t
+end
+
+init_magnetization(M, M0) = M0
+init_magnetization(M, M0::Real) = begin
+   x = similar(M, size(M,1))
+   x .= M0
+   return x
+end
```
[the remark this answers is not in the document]
pvillacorta: Yes, in fact, this function is no longer necessary... It was while I was creating the functions, but in the end I did it in a way that it was no longer necessary.

cncastillo:
I prefer to have a separate method for  `plot_phantom_map` if `Phantom{Motion}` has motion, so we don't loose the slider. 

There will be code repetition, unless we refactor the functions, or make the slider work. For now this is the easiest way, we can clean it up later.

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -0,0 +1,44 @@
...
+   mask_end = 1 .- vec(any(mask .== 0; dims=2))
+   if size(M, 2) > 1
+      M .*= mask
+      M .+= M0 .* (1 .- mask)
+   else
+      M .*= mask_end
+      M .+= M0 .* (1 .- mask_end)
+   end
+   return nothing
+end
```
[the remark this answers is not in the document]
cncastillo: Please try to offer a solution instead of saying you don't know how to do something. I don't have a lot of time to think about this for you.

Conceptually, there shouldn't be any difference between the two cases, and the code needs to reflect that. If there is a difference, there is a design problem and you are mixing two cases together that should be separated, like applying this function to a `Mag` vs `AbstractArray`.

This whole code snippet needs to be simplified.
```julia
mask .= (cumsum(mask; dims=2) .== 0)
mask_end = 1 .- vec(any(mask .== 0; dims=2))
if size(M, 2) > 1
  M .*= mask
  M .+= M0 .* (1 .- mask)
else
  M .*= mask_end
  M .+= M0 .* (1 .- mask_end)
end
```
Regarding the `cumsum`,  instead, you can use `accumulate` with a boolean "or."

```julia
julia> A = (rand(3, 4) .> 0.85)
3×4 BitMatrix:
 0  0  1  0
 1  0  0  0
 0  0  0  1

julia> accumulate(|, A; dims=2)
3×4 BitMatrix:
 0  0  1  1
 1  1  1  1
 0  0  0  1
```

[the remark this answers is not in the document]
cncastillo: Note that `reduce(+, ...)` and `accumulate(+, ...)` are the same as `sum` and `cumsum`.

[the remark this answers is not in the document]
pvillacorta: Except for the aspect of the dispatch between the magnetization and the vector, which can be applied, I have not been able to implement the rest of the changes. The accumulate function returns a bitmatrix that cannot be to handled on the GPU. Additionally, I tried to apply the approach of calculating t_reset, but for that, we need either the accumulate function we discussed or perform the interpolation.

So, although it may seem excesive, I don’t see a simpler option than interpolating, and at the same time, one that is compatible with both CPU and GPU.

Please keep in mind that I only turn to you after I have researched and tried to solve and address the problems on my own. The solution I end up committing is not the first one that comes to mind, but rather the result of many attempts and having tried it in various ways.

I will upload the suggested changes by removing the code snippet, but providing two definitions for the function instead.

KomaMRIBase/src/motion/motionlist/actions/arbitraryactions/FlowPath.jl
```diff
@@ -0,0 +1,37 @@
...
+           dz=[0.03 0.04; 0.04 -0.04],
+           spin_reset=[false false; false true]
+       )
+```
+"""
+@with_kw struct FlowPath{T<:Real} <: ArbitraryAction{T}
+    dx::AbstractArray{T}
+    dy::AbstractArray{T}
+    dz::AbstractArray{T}
+    spin_reset::AbstractArray{T}
```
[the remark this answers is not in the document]
pvillacorta: I have made some modifications to ArbitraryAction.jl to allow the coefficients and nodes to have different eltypes. It works on CPU and CUDA but I have to check in oneAPI because I don't trust it.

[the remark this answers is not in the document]
cncastillo: Thanks for looking into this, let's keep it like this, maybe add a TODO or open an issue to remember to change this at some point.

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -0,0 +1,44 @@
...
+
+init_time(t, seq_t, add_t0) = t
+init_time(t, seq_t::AbstractArray, add_t0) = begin
+   t1 = @view(seq_t[1])
+   return add_t0 ? [t1 (t1 .+ t)] : t1 .+ t
+end
```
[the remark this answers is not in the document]
pvillacorta: We could recreate with DiffEquations a scenario where a single spin jumps and resets its magnetization state. This could be done by solving two different ODE problems, but following a continuity with the times, and then adding the two results, right? I could try to do this but, maybe after merging?

KomaMRICore/test/runtests.jl
```diff
@@ -458,3 +415,91 @@ end
...
+
+                ## Solving using DiffEquations.jl
+                function bloch!(dm, m, p, t)
+                    mx, my, mz = m
+                    bx, by, bz = [B1e(t) * cos(φ), B1e(t) * sin(φ), (x(t) * Gx + y(t) * Gy + z(t) * Gz)]
+                    dm[1] = -mx / T2 + γ * bz * my - γ * by * mz
+                    dm[2] = -γ * bz * mx - my / T2 + γ * bx * mz
+                    dm[3] =  γ * by * mx - γ * bx * my - mz / T1 + M0 / T1
+                    return nothing
+                end
```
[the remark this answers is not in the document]
pvillacorta: Done!

cncastillo:
CUDA seems to be segfaulting on BuildKite, but I don't think that is our fault, so I re-run that job to see if it passes, also for Github's CI (Julia 1 / Ubuntu).

cncastillo:
@pvillacorta The CI is saying this

```julia
┌ KomaMRICore
│  WARNING: method definition for #outflow_spin_reset!#54 at /var/lib/buildkite-agent/builds/gpuci-15/julialang/komamri-dot-jl/KomaMRICore/src/simulation/Flow.jl:18 declares type variable T but does not use it.
└  
```

pvillacorta:
Yep, I removed it in the commits where I tried to make `spin_reset` a Matrix{Bool}, but I am not pushing them and I forgot to change that.

KomaMRIBase/src/motion/motionlist/actions/arbitraryactions/FlowPath.jl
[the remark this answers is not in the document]
pvillacorta: I open an issue so I don´t make more commits haha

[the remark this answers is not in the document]
rkierulf: Chiming in since I have an idea on how to fix the oneAPI cumsum kernel. Does this work?

```julia
@kernel function naive_cumsum!(B, @Const(A))
    i = @index(Global)
    T = eltype(A)

    cur_val = zero(T)
    for k ∈ 1:size(A, 2)
        @inbounds cur_val += A[i, k]
        @inbounds B[i, k] = cur_val
    end
end
```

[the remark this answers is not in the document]
pvillacorta: It works! Thank you, I have committed that change in [5780c31](https://github.com/JuliaHealth/KomaMRI.jl/pull/442/commits/5780c3102174e430393cb4d082fed23ab48b4b60)

cncastillo:
It looks like a proper error, but I will re-run it.

examples/3.tutorials/lit-05-SimpleMotion.jl
```diff
@@ -71,7 +67,7 @@ p2 = plot_image(abs.(image1[:, :, 1]); height=400) # hide
 # S_{\mathrm{MC}}\left(t\right)=S\left(t\right)\cdot\mathrm{e}^{\mathrm{i}\Delta\phi_{\mathrm{corr}}}=S\left(t\right)\cdot\mathrm{e}^{\mathrm{i}2\pi\boldsymbol{k}\left(t\right)\cdot\boldsymbol{u}\left(t\right)}
 # ```
 
-# In practice, we would need to estimate or measure the motion before performing a motion-corrected reconstruction, but for this example, we will directly use the displacement functions ``\boldsymbol{u}(\boldsymbol{x}, t)`` defined by `obj.motion::SimpleMotion`. 
+# In practice, we would need to estimate or measure the motion before performing a motion-corrected reconstruction, but for this example, we will directly use the displacement functions ``\boldsymbol{u}(\boldsymbol{x}, t)`` defined by `obj.motion::SimpleAction`.
```
cncastillo, line 70: `obj.motion::SimpleAction` this is not correct, it is a motion list with a simple action inside.

examples/3.tutorials/lit-05-SimpleMotion.jl
```diff
@@ -4,23 +4,19 @@ using KomaMRI # hide
 sys = Scanner() # hide
 
 # It can also be interesting to see the effect of the patient's motion during an MRI scan.
-# For this, Koma provides the ability to add `motion <: MotionModel` to the phantom.
-# In this tutorial, we will show how to add a [`SimpleMotion`](@ref) model to a 2D brain phantom.
+# For this, Koma provides the ability to add `motion <: AbstractMotionSet` to the phantom.
```
cncastillo, line 7: `motion <: AbstractMotionSet` I thought we changed `AbstractMotionSet` to `AbstractMotion`. Set has a precise mathematical meaning.

`AbstractMotion`:
- `struct NoMotion`
- `struct MotionList` (The same as an AbstractArray{Motion} but we didn't use arrays, because ... ? we will need to discuss this in the paper)
	- `struct Motion` (This one makes sense)
		- `Action <: AbstractAction`
		- `TimeSpan <: AbstractTimeSpan`
		- `SpinSpan <: AbstractSpinSpan` (Is the user meant to extend this one?)
		
Is it like this? 

What function must I define to generate a new `AbstractMotion`/`AbstractAction`/...?

If I want to define my new custom motion, should create a new AbstractAction or a new AbstractMotion? 

Documenting all of this is important, let's generate an issue.

KomaMRICore/src/simulation/Functors.jl
```diff
@@ -99,19 +97,23 @@ See also [`f32`](@ref).
 """
 f64(m) = paramtype(Float64, m)
 
+# Koma motion-related adapts
+adapt_storage(backend::KA.GPU, xs::MotionList) = MotionList(gpu.(xs.motions, Ref(backend)))
+adapt_storage(T::Type{<:Real}, xs::NoMotion) = NoMotion{T}()
+adapt_storage(T::Type{<:Real}, xs::MotionList) = MotionList(paramtype.(T, xs.motions))
+adapt_storage(T::Type{<:Real}, xs::Motion) = Motion(paramtype(T, xs.action), paramtype(T, xs.time), xs.spins)
```
cncastillo, line 104: Is this necessary? What happened with the ranges/vector?

cncastillo (changes requested):

KomaMRICore/src/simulation/Functors.jl
pvillacorta: The first `adapt` to GPU is indeed necessary. If I don't use it, the function |> gpu will try to convert the MotionList.motions attribute (which is a `Vector{Motion}`) to a `GPUArray{Motion}`, and we don't want that. That is also why I am wrapping the motion array inside the `MotionList` struct.
I will check for the other three adapts.

examples/3.tutorials/lit-05-SimpleMotion.jl
pvillacorta: I change `AbstractMotionSet` to `AbstractMotion` and put your message in issue https://github.com/JuliaHealth/KomaMRI.jl/issues/395

KomaMRICore/src/simulation/Functors.jl
cncastillo: What about using `gpu` instead of `adapt`?

pvillacorta: The `adapt` function does not make use of the `fmap` function, so it would not adapt recursively.
Concerning the necessity of the MotionList, I also think that there must be a way to omit it, and just let everything as it was initially stated in https://github.com/JuliaHealth/KomaMRI.jl/issues/410.
The main problem of this was that doing `obj.motion |> gpu` tried to convert that `Vector{Motion}`) to a `GPUArray{Motion}`, and if I defined a new adapt method like: 
`adapt_storage(backend::KA.GPU, xs::Vector{Motion}) = gpu.(backend, xs)`
it produced a conflict with other defined methods. We can see this later, as it maybe has an easy solution that I am not being able to see.

examples/3.tutorials/lit-05-SimpleMotion.jl
cncastillo: Ok thanks

> If I don't want to define a list, can I just define obj.motion = Translate(...)?

It would be nice if we could do this (motion without list)..maybe this can be a different issue.

KomaMRICore/src/simulation/Functors.jl
pvillacorta: As it is now, I can only omit the 4th line: the adapt to `Motion`

cncastillo: Ok, should I wait for the commit?

pvillacorta: e416800d1165086a399412f8c4ea41b58099f1d3

cncastillo: Oh ok, you did both in one commit, I just saw the AbstractMotion part.

cncastillo (approved):

cncastillo:
I am a little worried about this @pvillacorta 

MRI Lab/Bloch/CPU/8 thread(s) 	143972736.5 ns 	90572834 ns 	**1.59**

There seems to be a performance regression. I will merge it anyway, but if this is still the case, I would need to ask you to investigate.

PR #443, opened 2024-07-11
Optimize run_spin_precession! and run_spin_excitation! for CPU

rkierulf:
I've updated the run_spin_precession function to no longer use a matrix of size nspins x seq.Δt, in order to reduce memory use and (hopefully) simulation time as well. Instead, it now allocates arrays of size 1 x nspins, and steps through time to update them. The arrays are also pre-allocated at the beginning of run_sim_time_iter, so they can be re-used from block-to-block. 

Interestingly, running the MRI Lab benchmark on my computer, I did see significantly faster simulation times on the CPU (roughly 4x faster), and a ~80% decrease in memory usage, however, GPU performance for Metal was much slower, going from ~6 seconds to run the MRI Lab benchmark to ~46 seconds. I haven't looked deeply into why this is yet, but I suspect this approach of stepping through time is more suited to the CPU than GPU. When I write a new BlochKernel simulation method using KernelAbstractions, I think we should switch back to the matrix approach, so we will have one method specialized for the CPU and another for the GPU. For the GPU method, we can  have one thread for each entry in the nspins x seq.Δt matrix and compute the effective field all at once. There are parts that need to be sequential in time, but I think we can reduce across the time dimension in O(log(seq.Δt)) cycles. 

I was originally planning to have a separate pull request for optimizing run_spin_excitation, but I think it's simpler to just include in this pull request and merge together, so I'll work on that next. Other issues fixed:

- ~~In Benchmark.yml, the job is not run for external pull requests since the Buildkite action is not run~~
- If the user tries to run on the GPU but forgot to load the GPU backends, it will run on the CPU with Threads.nthreads(), not just 1 thread
- If the user tries to run on the CPU and Threads.nthreads()=1, they see the message: "Simulation will be run on the CPU with only 1 thread. To enable multi-threading, start julia with --threads=auto."


KomaMRICore/Project.toml
```diff
@@ -11,6 +11,7 @@ KomaMRIBase = "d0bc0b20-b151-4d03-b2a4-6ca51751cb9c"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
```
cncastillo, line 14:
```suggestion
```

KomaMRICore/ext/KomaAMDGPUExt.jl
```diff
@@ -1,6 +1,7 @@
 module KomaAMDGPUExt
 
 using AMDGPU
+using Suppressor
```
cncastillo, line 4:
```suggestion
```

KomaMRICore/ext/KomaAMDGPUExt.jl
```diff
@@ -20,6 +21,7 @@ end
 
 function __init__()
     push!(KomaMRICore.LOADED_BACKENDS[], ROCBackend())
+    @suppress AMDGPU.allowscalar(true)
```
cncastillo, line 24:
```suggestion
```

KomaMRICore/ext/KomaCUDAExt.jl
```diff
@@ -1,6 +1,7 @@
 module KomaCUDAExt
 
 using CUDA
+using Suppressor
```
cncastillo, line 4:
```suggestion
```

KomaMRICore/ext/KomaCUDAExt.jl
```diff
@@ -19,6 +20,7 @@ end
 
 function __init__()
     push!(KomaMRICore.LOADED_BACKENDS[], CUDABackend())
+    @suppress CUDA.allowscalar(true)
```
cncastillo, line 23:
```suggestion
```

KomaMRICore/ext/KomaMetalExt.jl
```diff
@@ -3,6 +3,7 @@
 module KomaMetalExt
 
 using Metal
+using Suppressor
```
cncastillo, line 6:
```suggestion
```

KomaMRICore/ext/KomaMetalExt.jl
```diff
@@ -16,16 +17,9 @@ function KomaMRICore._print_devices(::MetalBackend)
...
-#Once run_spin_excitation! and run_spin_precession! are kernel-based, this code
-#can be removed
-Base.cumsum(x::MtlVector{T}) where T = convert(MtlVector{T}, cumsum(KomaMRICore.cpu(x)))
-Base.cumsum(x::MtlArray{T}; dims) where T = convert(MtlArray{T}, cumsum(KomaMRICore.cpu(x), dims=dims))
-Base.findall(x::MtlVector{Bool}) = convert(MtlVector, findall(KomaMRICore.cpu(x)))
-
 function __init__()
     push!(KomaMRICore.LOADED_BACKENDS[], MetalBackend())
-    @warn "Metal does not support all array operations used by KomaMRI (https://github.com/JuliaGPU/Metal.jl/issues/348). GPU performance may be slower than expected"
+    @suppress Metal.allowscalar(true)
```
cncastillo, line 22:
```suggestion
```

KomaMRICore/ext/KomaoneAPIExt.jl
```diff
@@ -1,6 +1,7 @@
 module KomaoneAPIExt
 
 using oneAPI
+using Suppressor
```
cncastillo, line 4:
```suggestion
```

KomaMRICore/ext/KomaoneAPIExt.jl
```diff
@@ -17,15 +18,9 @@ function KomaMRICore._print_devices(::oneAPIBackend)
...
-#Temporary workaround since oneAPI.jl (similar to Metal) does not support some array operations
-#Once run_spin_excitation! and run_spin_precession! are kernel-based, this code can be removed
-Base.cumsum(x::oneVector{T}) where T = convert(oneVector{T}, cumsum(KomaMRICore.cpu(x)))
-Base.cumsum(x::oneArray{T}; dims) where T = convert(oneArray{T}, cumsum(KomaMRICore.cpu(x), dims=dims))
-Base.findall(x::oneVector{Bool}) = convert(oneVector, findall(KomaMRICore.cpu(x)))
-
 function __init__()
     push!(KomaMRICore.LOADED_BACKENDS[], oneAPIBackend())
-    @warn "oneAPI does not support all array operations used by KomaMRI. GPU performance may be slower than expected"
+    @suppress oneAPI.allowscalar(true)
```
cncastillo, line 23:
```suggestion
```

cncastillo (commented):
Generally, allowing scalar indexing makes the GPU suffer.

rkierulf:
When stepping through the sequence in time, we do need to access arrays such as seq.Gx / Gy / Gz, seq.ADC, and sig at a specific index for each time step, hence the scalar indexing. My understanding from https://discourse.julialang.org/t/overcoming-slow-scalar-operations-on-gpu-arrays/49554/5 is that this is ok since the number of spins is expected to be >> than the length of seq.Δt inside each block, and nspins is the dimension of the array we are broadcasting along. Although, as I mentioned earlier, the for-loop stepping through time approach is probably inefficient in general for the GPU, and we should change back to using a matrix for a future BlochKernel sim method. 

At one point, I did have it so that each scalar index was turned into an array and broadcasted, so for example:

Bz_2 .= x .* seq.Gx[seq_idx,:] .+ y .* seq.Gy[seq_idx,:] .+ z.* seq.Gz[seq_idx,:] .+ p.Δw / T(2π * γ)

instead of:

Bz_2 .= x .* seq.Gx[seq_idx] .+ y .* seq.Gy[seq_idx] .+ z.* seq.Gz[seq_idx] .+ p.Δw / T(2π * γ)

But this was slightly slower (both on the CPU and GPU) compared with just using scalar indexing, so I changed it.

rkierulf:
Another possible approach for run_spin_precession would be to still use the matrix-based approach, but pre-allocate the matrix with size nspins x seq.Δt for the longest block in the sequence, and then for blocks with a smaller seq.Δt use a view of this matrix. This would still cut memory usage down by a lot I think. 

I'm pretty sure the main reason for the time being faster on the CPU is that we are only calculating Mxy if seq.ADC is true for that time step, whereas before we were calculating it for every time step, and this is the most expensive calculation. We could still do this using a matrix, however. I can play around a bit with this approach and see how the time compares.

cncastillo:
Yeah, I'm not sure what's best :/. Let's discuss it tomorrow!

cncastillo:
Yeah, I'm not sure what's best :/. Let's discuss it tomorrow!

rkierulf:
I've updated the file structure of KomaMRICore to match what we discussed. The default run_spin_precession! and run_spin_excitation! functions are now in a new file SimulationMethod.jl and are dispatched at the most abstract level for the sim_method, backend, and prealloc parameters. The specialized CPU implementation for run_spin_precession! is in a separate file BlochCPU.jl and is dispatched to for sim_method::Bloch, backend::KA.CPU, and prealloc::BlochCPUPrealloc. I also added an optimized implementation for run_spin_excitation! which improved memory usage a bit more for the MRI Lab benchmark but didn't seem to affect the simulation time.

Let me know if there's any more changes you want made before merging!

rkierulf:
Doing some more profiling this morning, I was surprised to see some lines allocating more than expected. It turns out that there were quite a few statements that were not fully broadcasted. Apparently, even if multiplying / dividing by a scalar, it still helps to use ./ or .* compared with / or *. I think this makes it easier for the compiler broadcast the entire statement. Making these changes, I was able to get memory usage to decrease by a factor of ~15 from where it was immediately before! However, simulation time only decreased by about 10%, so that at this point, we are clearly compute-bound and not memory-bound.

cncastillo:
Hi, it looks good! 

There's a few details it would be nice to change before merging 

(1) Can we call the previous Bloch method for `::SimulationMethod` (fallback) but also for `::BlochSimple`? And move it to its own folder for discover ability.

This is not for now, but:

(2) To organize the pre allocations, maybe we can put them in structs. For example, I think it makes sense to have
- M::Mag (or another spin state)
    - M.xy complex array Nspins x 1
    - M.z   real array Nspins x 1
- B_new::Mag
    - B_new.xy complex array Nspins x 1
    - B_new.z   real array Nspins x 1
- B_old::Mag
    - B_old.xy complex array Nspins x 1
    - B_old.z   real array Nspins x 1
- Rot::Spinor
    - Rot.alpha complex array Nspins x 1
    - Rot.beta   complex array Nspins x 1
So all variables have a physical meaning, basically we're storing the current `SimulationState`. 

For the excitation for now we only use B_new (I would like to add a trapezoidal integration later). 

For the precession, the term exp/cis(phi) I think is the same as the Spinor alpha, with beta=0.

Mag is not a very good name for the struct (as it could also be used for B_new and B_old), it is a vector in SU2 representation, VectorSU2.

https://onlinelibrary.wiley.com/doi/pdf/10.1002/9783527630097.app1#:~:text=We%20define%20a%20spinor%20as,two%2D%20dimensional%20complex%20variable%20space.

cncastillo:
Just waiting for the CI to pass to merge!

PR #445, opened 2024-07-12
Trying to fix benchmarking for PRs

cncastillo:


cncastillo:
Removing 
```
    branches:
      - master
```
It makes it so it doesn't crash ... but because it didn't run.

PR #466, opened 2024-08-20
Time Comparison

rkierulf:
As suggested by @cncastillo, I'm submitting a pull request reverting the repository to the first commit where the benchmarking CI was running (https://github.com/JuliaHealth/KomaMRI.jl/commit/5b6e9fbbda2311e57f3988185f1d3b13eefe3da1) so we can get an official time comparison by running the benchmarks again. We're curious if the AMD time improvements seen after (https://github.com/JuliaHealth/KomaMRI.jl/commit/1457a4c3ae1e3e7cc40819c534d8c36620bccb75) are actually due to the changes in that commit or an AMD change.


rkierulf:
When I merged, it looked like it ignored the git reset and only merged in the one commit that was ahead of the main branch. Just a second ...

PR #467, opened 2024-08-20
Time Comparison (Attempt 2)

rkierulf:
2nd attempt at reverting to the commit adding the benchmarking CI for a time comparison. I think it should work this time since I used this solution here: https://stackoverflow.com/questions/32841671/reverting-to-a-specific-commit-without-losing-history which shouldn't rewrite history like git reset.


cncastillo:
The `pull-requests: write` may be necessary for the benchmarking report, but let's see!

PR #468, opened 2024-08-22
Add Distributed Examples to Documentation

rkierulf:
I added two examples I tested yesterday to the "how-to" section of the documentation, one for running a simulation using multiple GPUs, and another using multiple nodes on a SLURM cluster. This isn't included in the example scripts, but I verified the accuracy compared to a normal simulation with these lines of code added at the end:

```julia
signal_2 = simulate(obj, seq, sys)
println([sum(abs.(signal.profiles[i].data .- signal_2.profiles[i].data)) / length(signal.profiles[i].data) for i=1:length(signal.profiles)])
```

The values I saw were all on the scale of 10e-6 / 10e-7, which is near machine precision for Float32s.


cncastillo:
https://juliahealth.org/KomaMRI.jl/previews/PR468/

cncastillo:
btw @rkierulf I think this could be very useful to simulate the "aorta" with flowing spins, as you can distribute the phantom to multiple nodes or GPUs!!

pvillacorta:
Nice! Once #442 is finished, I will try to simulate with the aorta and make and example :)

cncastillo (approved):
This is very nice! quite amazing that it works:
 - add a test for `+(sig1, sig2)`
 - rename the `sig`'s to `raw`'s
 - add the images
 - add the SLURM script as dropdowns with ```!!! details "SLURM script" ... ``` (example: https://github.com/JuliaHealth/KomaMRI.jl/blob/master/docs/src/index.md?plain=1)

cncastillo:
This is what the HPC example is actually doing

KomamultiNodeCPU

PR #484, opened 2024-09-22
Custom values for brain_phantom

gsahonero:
It would be useful to be able to change the default values. For example, when dealing with T1 or T2 changes due to different magnetic field strengths. 

The changes on `brain_phantom2D` and `brain_phantom3D` allow the user to define the desired values if passed as an argument.


cncastillo:
I think this is a good idea, but how this is implemented adds too many lines and complexity (you repeat the same code twice). 

I would suggest having default values in `custom_values` (`tissue_properties` I think it is a better name) that give exactly what we have now.

If the user specifies `tissue_properties`, you could merge the custom properties with the new ones so that the user can define just a few tissues. Maybe a dictionary would be suitable for this.

I think this could be done without adding so many lines.

cncastillo:
I think the default properties could do something similar to the simulation parameters, that have a default_brain_tissue_properties function. And the user defined parameters are merged to that. As it doesn't make sense to be defined twice (one in 2d and one in 3d).

Also is makes more sense that the keys for the dictionary were the tissue names, and that inside you can specify the per tissue properties in some documented order. As the other way around the user needs to remember which tissue was which and it's prone to errors.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -29,7 +31,7 @@ julia> obj = Phantom(x=[0.0])
 julia> obj.ρ
 ```
 """
-@with_kw mutable struct Phantom{T<:Real}
+@with_kw mutable struct Phantom{T<:Number}
```
cncastillo: Revert this as it is not relevant for this PR.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -48,6 +50,8 @@ julia> obj.ρ
     #Diff::Vector{DiffusionModel}  #Diffusion map
     #Motion
     motion::AbstractMotion{T} = NoMotion{eltype(x)}() 
+    Mxy::AbstractVector{ComplexF32} = zeros(eltype(x), size(x)) .+ 0im
+    Mz::AbstractVector{ComplexF32} = zeros(eltype(x), size(x)) .+ 0im
```
cncastillo: Revert this as it is not relevant for this PR.

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -985,6 +985,252 @@ function plot_kspace(seq::Sequence; width=nothing, height=nothing, darkmode=fals
...
+    end
+    if width !== nothing
+        l.width = width
+    end
+
+    return Plot(trace, l)
+end
+
+
+
```
cncastillo: Revert these changes, as they are not relevant to this PR.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +549,107 @@ function check_phantom_arguments(nd, ss, us)
...
+- `T1, T2, T2s, ρ, Δw`: (`::Matrix`) matrices of the same size of class with the tissues properties information
+
+# Examples
+```julia-repl
+julia> T1, T2, T2s, ρ, Δw = default_brain_tissue_properties(class, tissue_properties)
+
+julia> T1, T2, T2s, ρ, Δw = default_brain_tissue_properties(class)
+```
+"""
+function default_brain_tissue_properties(class, tissue_properties = nothing)
```
cncastillo, line 570: Change `class` with `labels` as it is more obvious what it means.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +549,107 @@ function check_phantom_arguments(nd, ss, us)
...
+            "FAT1" => [350, 70, 58, 1, -220],
+            "MUSCLE" => [900, 47, 30, 1, 0],
+            "SKIN/MUSCLE" => [569, 329, 58, 1, 0],
+            "SKULL" => [0, 0, 0, 0, 0],
+            "VESSELS" => [0, 0, 0, 0, 0],
+            "FAT2" => [500, 70, 61, 0.77, -220],
+            "DURA" => [2569, 329, 58, 1, 0],
+            "MARROW" => [500, 70, 61, 0.77, 0])
+        # Define spin property vectors
+    end
```
cncastillo, line 587: It is better to `merge` the `tissue_properties` (user defined) with `default_tissue_properties` instead of adding an `if isnothing`.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -212,7 +217,7 @@ Default ss=4 sample spacing is 2 mm. Original file (ss=1) sample spacing is .5 m
 - `axis`: (`::String`, `="axial"`, opts=[`"axial"`, `"coronal"`, `"sagittal"`]) orientation of the phantom
 - `ss`: (`::Integer or ::Vector{Integer}`, `=4`) subsampling parameter for all axes if scaler, per axis if 2 element vector [ssx, ssy]
 - `us`: (`::Integer or ::Vector{Integer}`, `=1`)  upsampling parameter for all axes if scaler, per axis if 2 element vector [usx, usy], if used ss is set to ss=1
-
+- `tissue_properties`: (`::Dict`, `=nothing`) phantom tissue properties in ms and Hz considering the available tissues
```
cncastillo, line 216: This breaks the consistency of units in Koma, it is better to use SI units. All units are specified as SI, but they are displayed differently.

KomaMRICore/src/simulation/SimulatorCore.jl
```diff
@@ -409,6 +409,8 @@ function simulate(
         out = Xt
     elseif sim_params["return_type"] == "mat"
         out = sig
+    elseif sim_params["return_type"] == "full"
+        out = Dict("state" => Xt, "mat" => sig)
```
cncastillo: I like this change (I would prefer to output a Tuple), but it is not relevant to this PR.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -223,11 +228,28 @@ julia> obj = brain_phantom2D(; axis="sagittal", ss=1)
...
+        "WM" => [746, 70, 61, 0.77, 0],
+        "FAT1" => [0, 0, 0, 0, 0],
+        "MUSCLE" => [0, 0, 0, 0, 0],
+        "SKIN/MUSCLE" => [0, 0, 0, 0, 0],
+        "SKULL" => [0, 0, 0, 0, 0],
+        "VESSELS" => [0, 0, 0, 0, 0],
+        "FAT2" => [0, 0, 0, 0, 0],
+        "DURA" => [0, 0, 0, 0, 0],
+        "MARROW" => [0, 0, 0, 0, 0])
+julia> obj = brain_phantom2D(; tissue_properties=phantom_values)
```
cncastillo, line 241: Add spaces so the columns are aligned.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +549,107 @@ function check_phantom_arguments(nd, ss, us)
...
+        (class .== 93) * tissue_properties["FAT1"][5] .+ #FAT1
+        (class .== 116) * tissue_properties["MUSCLE"][5] .+ #MUSCLE
+        (class .== 139) * tissue_properties["SKIN/MUSCLE"][5] .+ #SKIN/MUSCLE
+        (class .== 162) * tissue_properties["SKULL"][5] .+ #SKULL
+        (class .== 185) * tissue_properties["VESSELS"][5] .+ #VESSELS
+        (class .== 209) * tissue_properties["FAT2"][5] .+ #FAT2
+        (class .== 232) * tissue_properties["DURA"][5] .+ #DURA
+        (class .== 255) * tissue_properties["MARROW"][5] #MARROW
+    return T1, T2, T2s, ρ, Δw
+end
```
cncastillo, line 651: This could be done in a for loop, so it outputs a Tuple, and to avoid writing the same thing 5 times only changing a number.

cncastillo (changes requested):

gsahonero:
Sorry about those last commits, the last one is the good one. 

In summary, I wrote a `default_brain_tissue_properties` that is being called for creating the 2D and 3D brain phantom. That way, redefinition is avoided. Also, I changed the definition of tissue properties and now the names are being used. I hope those changes may address the suggestions.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +539,62 @@ function check_phantom_arguments(nd, ss, us)
...
+        "MUSCLE" => [0.900, 0.047, 0.030, 1, 0],
+        "SKIN/MUSCLE" => [0.569, 0.329, 0.058, 1, 0],
+        "SKULL" => [0, 0, 0, 0, 0],
+        "VESSELS" => [0, 0, 0, 0, 0],
+        "FAT2" => [0.500, 0.070, 0.061, 0.77, -3.84], #-220 Hz
+        "DURA" => [2.569, 0.329, 0.058, 1, 0],
+        "MARROW" => [0.500, 0.070, 0.061, 0.77, 0])
+    
+    tissue_properties = merge(default_properties, tissue_properties)
+    flat_labels = reshape(labels, :)
```
cncastillo, line 581:
```suggestion
    flat_labels = labels[:]
```

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +539,62 @@ function check_phantom_arguments(nd, ss, us)
...
+        (flat_labels .== 139) * tissue_properties["SKIN/MUSCLE"][i] .+ #SKIN/MUSCLE
+        (flat_labels .== 162) * tissue_properties["SKULL"][i] .+ #SKULL
+        (flat_labels .== 185) * tissue_properties["VESSELS"][i] .+ #VESSELS
+        (flat_labels .== 209) * tissue_properties["FAT2"][i] .+ #FAT2
+        (flat_labels .== 232) * tissue_properties["DURA"][i] .+ #DURA
+        (flat_labels .== 255) * tissue_properties["MARROW"][i] #MARROW
+    end
+    properties = reshape(properties, (size(labels)..., 5))
+
+    return selectdim(properties, ndims(properties), 1), selectdim(properties, ndims(properties), 2), selectdim(properties, ndims(properties), 3), selectdim(properties, ndims(properties), 4), selectdim(properties, ndims(properties), 5)
```
cncastillo, line 599: This is overly complicated.

```julia
return (properties[:, i] for i in 1:Nproperties)
```

cncastillo (changes requested):

KomaMRIBase/src/datatypes/Phantom.jl
gsahonero: I tried that approach. However, with that idea, a flat vector will be returned and a reshape should be done on the `brain_phantom_2D` function. Using `selectdim` generalizes for either 2D or 3D implementations and no reshape is required twice on the 2D and 3D functions.

cncastillo:
It looks a lot better now!

cncastillo:
Ok I had to merge another branch before and now this has merge conflicts. Could you solved them so I can merge this?

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -223,86 +224,49 @@ julia> obj = brain_phantom2D(; axis="sagittal", ss=1)
...
+        "MARROW" => [0, 0, 0, 0, 0])
+julia> obj = brain_phantom2D(; tissue_properties=phantom_values)
+
 julia> plot_phantom_map(obj, :ρ)
 ```
 """
-function brain_phantom2D(; axis="axial", ss=4, us=1)
+function brain_phantom2D(; axis="axial", ss=4, us=1, tissue_properties = Dict())
     # check and filter input    
+    # check more spins
```
cncastillo, line 248:
```suggestion
```

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -195,6 +195,7 @@ function heart_phantom(;
     return phantom
 end
 
+
```
cncastillo, line 198:
```suggestion
```

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -482,7 +409,7 @@ function pelvis_phantom2D(; ss=4, us=1)
 
     # subsample or upsample the phantom data
     class = repeat(data["pelvis3D_slice"][1:ssx:end, 1:ssy:end]; inner=[usx, usy])
-
+
```
cncastillo, line 412:
```suggestion

```

cncastillo (changes requested):
Nice. Apply these last minor suggestions, and I will merge them (some spaces and new lines are being added).

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +535,61 @@ function check_phantom_arguments(nd, ss, us)
...
+        "GM" => [0.833, 0.083, 0.069, 0.86, 0],
+        "WM" => [0.500, 0.070, 0.061, 0.77, 0],
+        "FAT1" => [0.350, 0.070, 0.058, 1, -3.84], #-220 Hz
+        "MUSCLE" => [0.900, 0.047, 0.030, 1, 0],
+        "SKIN/MUSCLE" => [0.569, 0.329, 0.058, 1, 0],
+        "SKULL" => [0, 0, 0, 0, 0],
+        "VESSELS" => [0, 0, 0, 0, 0],
+        "FAT2" => [0.500, 0.070, 0.061, 0.77, -3.84], #-220 Hz
+        "DURA" => [2.569, 0.329, 0.058, 1, 0],
+        "MARROW" => [0.500, 0.070, 0.061, 0.77, 0])
```
cncastillo, line 574: Improve indentation, and reorder properties (not done here):
```suggestion
    default_properties = Dict(
        #            T1      T2     T2*     ρ     Δw
        "CSF"         => [2.569, 0.329, 0.058, 1,    0],
        "GM"          => [0.833, 0.083, 0.069, 0.86, 0],
        "WM"          => [0.500, 0.070, 0.061, 0.77, 0],
        "FAT1"        => [0.350, 0.070, 0.058, 1,   -3.84], # -220 Hz
        "MUSCLE"      => [0.900, 0.047, 0.030, 1,    0],
        "SKIN/MUSCLE" => [0.569, 0.329, 0.058, 1,    0],
        "SKULL"       => [0,     0,     0,     0,    0],
        "VESSELS"     => [0,     0,     0,     0,    0],
        "FAT2"        => [0.500, 0.070, 0.061, 0.77, -3.84], # -220 Hz
        "DURA"        => [2.569, 0.329, 0.058, 1,    0],
        "MARROW"      => [0.500, 0.070, 0.061, 0.77, 0]
    )
```

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +535,61 @@ function check_phantom_arguments(nd, ss, us)
...
+        (labels .== 93) * tissue_properties["FAT1"][i] .+ #FAT1
+        (labels .== 116) * tissue_properties["MUSCLE"][i] .+ #MUSCLE
+        (labels .== 139) * tissue_properties["SKIN/MUSCLE"][i] .+ #SKIN/MUSCLE
+        (labels .== 162) * tissue_properties["SKULL"][i] .+ #SKULL
+        (labels .== 185) * tissue_properties["VESSELS"][i] .+ #VESSELS
+        (labels .== 209) * tissue_properties["FAT2"][i] .+ #FAT2
+        (labels .== 232) * tissue_properties["DURA"][i] .+ #DURA
+        (labels .== 255) * tissue_properties["MARROW"][i] #MARROW
+        push!(properties, temp)
+    end
```
cncastillo, line 592: Just realized this is iterating for (T1, T2, T2*, ρ, Δw) instead of the tissue properties. This code has many problems:

- for loop is hard-coded to `i = 1:5`, what is 5? what if we want more properties? if we change it to 6 it will error.
- the properties are not preallocated, you `push!` them.
- The order of the properties does not make much sense. T1 and T2 before ρ? 
- The Δw = -3.84 doesn't seem correct. 

Suggestion:
```julia
Nproperties = 5 # ρ, T1, T2, T2*, Δw, explicit ... 5 is the number of properties
# or
Nproperties = size(tissue_properties ... , 2) # Or something like that
properties = zeros(size(...), Nproperties)
tissue_labels = [23, 46, ....]
for (label, tissue) in zip(tissue_labels, tissue_properties)
	for property in 1:Nproperties
		properties[:,:, property] .+= (labels .== label) .* tissue[property]
	end
end
```

cncastillo:
Hi @gsahonero when do you think you could finish this?

gsahonero:
Sorry for taking this long. I hope the changes address the requested modifications.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +535,59 @@ function check_phantom_arguments(nd, ss, us)
...
+        "VESSELS"       => [0,      0,      0,      0,      0],
+        "FAT2"          => [0.77,   0.500,  0.070,  0.061,  -220*2π], #-220 Hz
+        "DURA"          => [1,      2.569,  0.329,  0.058,  0],
+        "MARROW"        => [0.77,   0.500,  0.070,  0.061,  0])
+    
+    tissue_properties = merge(default_properties, tissue_properties)
+    props = ["ρ", "T1", "T2", "T2s", "Δw"]
+    Nproperties = length(props) 
+    tissue_labels = [23, 46, 70, 93, 116, 139, 162, 185, 209, 232, 255]
+    tissue_texts = ["CSF", "GM", "WM", "FAT1", "MUSCLE", "SKIN/MUSCLE", "SKULL", "VESSELS", "FAT2", "DURA", "MARROW"]
```
cncastillo, line 580: You are defining this twice, use the keys of the `default_properties` dictionary, like in my example.

`data_properties[i, ntuple(j -> 1:dims[j+1], length(dims)-1)...]`  this is not very elegant, just do `data_properties[i, :, :, :]`, no need to define `dims` as the data is always three dimensional.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +535,59 @@ function check_phantom_arguments(nd, ss, us)
...
+    for i=1:Nproperties
+        for (label, tissue) in zip(tissue_labels, tissue_texts)
+            data_properties[i, ntuple(j -> 1:dims[j+1], length(dims)-1)...] += (labels .== label)*tissue_properties[tissue][i]
+        end
+    end
+    # To make it a tuple of five elements
+    properties = Tuple(data_properties[i, :, :, :] for i in 1:size(data_properties, 1))
+    # To avoid singleton dimensions
+    properties = Tuple(reshape(item, filter(d -> d>1, size(item))...) for item in properties)
+    return properties
```
cncastillo, line 592: I don't get why this is even necessary `return (data_properties[:, :, :, i] for i in 1:Nproperties)` should be enough. Iterating over `data[i,:,:,:]` instead of `data[:,:,:,i]` is a lot of generating unnecessary code.

cncastillo (changes requested):

KomaMRIBase/src/datatypes/Phantom.jl
gsahonero: Yeah, well, the problem is that the keys of the dictionary come in a different order, i.e., `keys(tissue_properties)` is not consistent with the creation order. 

Also, the problem with doing `data_properties[i, :, :, :]` is the inconsistent shape that `labels` has between 2D and 3D phantoms. The original code produces `labels` with 2 or 3 dimensions depending on what is being asked (`brain_phantom2D` or `brain_phantom3D` - which kind of makes sense but it also makes sense to introduce a singleton dimension on the `brain_phantom2D`'s labels to then generalize the code to 3D). In any case, this inconsistency made `data_properties[i, ntuple(j -> 1:dims[j+1], length(dims)-1)...]` required. 

Nevertheless, to don't use those lines, I modified the code.

KomaMRIBase/src/datatypes/Phantom.jl
gsahonero: Again, the problem was the shape inconsistency of the original implementation shape. In this case, `x` and `y` produced 2D matrices, while the `default_brain_tissue_properties` function returned tuples of 3D matrices (with a singleton dimension). It might seem unnecessary, but it was necessary to not modify other parts of the code. Because of that, it made sense to remove the singleton dimension using `Tuple(reshape(item, filter(d -> d>1, size(item))...) for item in properties)`. 

Nevertheless, to don't use those lines, I modified the code.

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -612,5 +539,56 @@ function check_phantom_arguments(nd, ss, us)
...
+        "VESSELS"       => [0,      0,      0,      0,      0],
+        "FAT2"          => [0.77,   0.500,  0.070,  0.061,  -220*2π], #-220 Hz
+        "DURA"          => [1,      2.569,  0.329,  0.058,  0],
+        "MARROW"        => [0.77,   0.500,  0.070,  0.061,  0])
+    
+    tissue_properties = merge(default_properties, tissue_properties)
+    props = ["ρ", "T1", "T2", "T2s", "Δw"]
+    Nproperties = length(props)
+    # Order: CSF, DURA, FAT1, FAT2, GM, MARROW, MUSCLE, SKIN/MUSCLE, SKULL, vESSELS, WM
+    tissue_labels = [23, 232, 93, 209, 46, 255, 116, 139, 162, 185, 70]
```
cncastillo, line 584: Could you try one last thing? 

The sorting and use of `tissue_texts`, does not make a lot of sense.  It is better if `tissue_labels` is also a dictionary (`Dict("CSF" => 23, ...)`), and we index it by the `default_properties` keys.

You should be able to do:

```julia
for i=1:Nproperties
    for (tissue, property) in tissue_properties
        data_properties[i, :, :, :] += (labels .== tissue_labels[tissue]) * property[i]
    end
end
```
Also, don't forget the spaces between the `*`!

cncastillo (approved):

KomaMRIBase/src/datatypes/Phantom.jl
gsahonero: Thank you for the suggestion! I implemented it in this last commit.

PR #485, opened 2024-09-24
Fix bug in `plot_phantom_map` when `length(obj)` > `max_spins`

pvillacorta:
I discovered this testing with big phantoms. I caused this error when dispatching `plot_phantom_map` for static and dynamic phantoms. 
I also add the description of some missing keyword arguments in the docstring


cncastillo:
Why was this not catched by the tests?

pvillacorta:
> Why was this not catched by the tests? 

The default `max_spins` value is 100_000, so this would not be noticeable unless you test with a phantom bigger than that. It was not the case for the tests

cncastillo:
Then we can put `max_spins = 1_000` in the tests, so that part of the code is being tested.

cncastillo (approved):

PR #486, opened 2024-09-26
Fix sum of `Grad`s

pvillacorta:
This is a temporary change to solve a bug when rotating sequences:

```julia-repl
julia> using KomaMRI
julia> seq = PulseDesigner.EPI_example()
julia> plot_seq(seq[1:10])
```
image

```julia-repl
julia> seq_r = rotz(π/2) * seq
julia> plot_seq(seq_r[1:10])
```
image

This bug happens when rotating a `Sequence`, as we are implicitly calling `+(Grad(0, 0), Grad(A, T))`, and this function assigns the `T`, `delay`, `rise` and `fall` of the first `Grad` to the result. A temporary solution is to assign the maximum of each one of these values:
```julia
+(x::Grad, y::Grad) = Grad(x.A .+ y.A, max(x.T, y.T), max(x.rise, y.rise), max(x.fall, y.fall), max(x.delay, y.delay)) 
```

But a better solution would be to "stack" the gradients if they have different time-related parameters. See https://github.com/JuliaHealth/KomaMRI.jl/issues/487


cncastillo (approved):
Looks good to me, thanks!

PR #488, opened 2024-09-27
Fix bugs related with `SpinRange` and flow

pvillacorta:
I discovered these bugs when testing with the artery.phantom (now deleted because of its size, I should upload it to another repository, Zenodo or as an artifact).

### DIFF
--- a/KomaMRIBase/src/motion/motionlist/Motion.jl
+++ b/KomaMRIBase/src/motion/motionlist/Motion.jl
@@ -164,11 +164,11 @@ Base.:(≈)(m1::Motion, m2::Motion)  = (typeof(m1) == typeof(m2)) & reduce(&, [g
 """ Motion sub-group """
 function Base.getindex(m::Motion, p)
     idx, spin_range = m.spins[p]
-    return Motion(m.action[idx], m.time, spin_range)
+    return idx !== nothing ? Motion(m.action[idx], m.time, spin_range) : nothing
 end
 function Base.view(m::Motion, p)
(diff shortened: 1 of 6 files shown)


cncastillo:
Can you explain why this check is needed?

```julia
function Base.getindex(m::Motion, p)
    idx, spin_range = m.spins[p]
    return idx !== nothing ? Motion(m.action[idx], m.time, spin_range) : nothing
end
```

pvillacorta:
> Can you explain why this check `idx !== nothing` is needed?
> 
> ```julia
> function Base.getindex(m::Motion, p)
>     idx, spin_range = m.spins[p]
>     return idx !== nothing ? Motion(m.action[idx], m.time, spin_range) : nothing
> end
> ```

I need this when, for example I have a `m=Motion` with `SpinRange(501:1000)` and I do `m[1:500]`, that is, when indexing with out-of-range indices. As the intersection between the two ranges is null, the return SpinRange would be `SpinRange(0:0)`, but that makes no sense. So, in that case, it is better to return `nothing`. 
Imagine that this motion is part of a phantom `obj` with 1000 spins:
```julia-repl
julia> obj.motion
MotionList(
   Translate(..., SpinRange(501:1000))
)
```

When indexing `obj[1:500]`: the resulting motion should be `NoMotion`
```julia-repl
julia> obj[1:500].motion
NoMotion(...)
```

cncastillo:
Ok!

cncastillo (approved):

Issue #498, opened 2024-10-16
Troubles defining Float32 or empty phantoms with KomaMRI 0.9


Tooine:
### What happened?

Hi, 

I switched to KomaMRI 0.9 and observed some performance improvements with simulations, which is great :+1: 

**Yet, I had some troubles (see below) with the way I define Float32 or empty phantoms with the new version.**
I am not sure if this is a bug or just the new expected behavior since 0.9, as for both situations, I found a (imo, less esthetic) way to define it as I want. 

Note that, **if this is not a bug, there might be parts of the documentation to modify**, like in the $T_2^*$ section of this [page](https://juliahealth.org/KomaMRI.jl/stable/tutorial-pluto/01-gradient-echo-spin-echo/), where the following empty phantom definition does not work anymore (at least for me)
```julia
# (3.1.1) Create an empty phantom
obj_t2star = Phantom{Float64}(x=[])
```

**Issue 1: Float32 vs Float64 phantom**
Creating a phantom encoded in Float32 like this leads to an error
```julia
Phantom{Float32}(x=[1.0])
→ ERROR: MethodError: Cannot `convert` an object of type NoMotion{Float64} to an object of type KomaMRIBase.AbstractMotion{Float32}
```
but this works fine and produces a phantom encoded in Float32 (including the motion!)
```julia
Phantom(x=Float32[1.0])
→ Phantom{Float32}
  name: String "spins"
  x: Array{Float32}((1,)) Float32[1.0]
  y: Array{Float32}((1,)) Float32[0.0]
  z: Array{Float32}((1,)) Float32[0.0]
  ρ: Array{Float32}((1,)) Float32[1.0]
  T1: Array{Float32}((1,)) Float32[1.0f6]
  T2: Array{Float32}((1,)) Float32[1.0f6]
  T2s: Array{Float32}((1,)) Float32[1.0f6]
  Δw: Array{Float32}((1,)) Float32[0.0]
  Dλ1: Array{Float32}((1,)) Float32[0.0]
  Dλ2: Array{Float32}((1,)) Float32[0.0]
  Dθ: Array{Float32}((1,)) Float32[0.0]
  motion: NoMotion{Float32} NoMotion{Float32}()
```

**Issue 2: Empty phantom**
Those commands do not work anymore
```julia
Phantom(x=[])
Phantom{Float32}(x=[])
Phantom{Float64}(x=[])
→ ERROR: MethodError: no method matching zero(::Type{Any})
```
[...]

PR #495, opened 2024-10-09
Phantom documentation

pvillacorta:
This PR aims to solve #395 and probably #394.

I am using Literate.jl to make all the new documents

https://juliahealth.org/KomaMRI.jl/previews/PR495/

To Do List: https://github.com/JuliaHealth/KomaMRI.jl/pull/495#issuecomment-2407634996

PD: it would be nice to merge #561 before this PR is merged.

PR #499, opened 2024-10-17
Fix Phantom-related bugs (#498 and more)

pvillacorta:
This PR solves the following:

- #498 
`Phantom(x=[])`, `Phantom{Float64}(x=[])` or `Phantom{Float32}(x=[])` are no longer supported*, since the element type of the Phantom is Inferred from its input arguments, and defining `x=[]` (eltype `Any`) is not a good practice in Julia.
Now, it is possible to define an empty Phantom this way:
    ```julia
    Phantom() # default eltype is Float64
    Phantom{Float64}()
    Phantom{Float32}()
    Phantom{Int64}()
    (...)
    ```
    *The example in [this page](https://juliahealth.org/KomaMRI.jl/stable/tutorial-pluto/01-gradient-echo-spin-echo/) is updated too

- [This bug](https://github.com/JuliaHealth/KomaMRI.jl/pull/495#issuecomment-2402722172) when creating a motionlist from an action.
- `plot_phantom_map`: now 2D or 1D phantoms in yz plane are displayed correctly. See [this comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/495#issuecomment-2413961563). Moreover, the prefix for the time slider in dynamic plots was incorrect ("x="). This has changed to be a ("t=").


pvillacorta:
Buildkite CI is still failing because of https://github.com/JuliaGPU/oneAPI.jl/issues/473

cncastillo:
Is there any other problem besides oneAPI.jl not working on Julia 1.11? We could merge this and register a new version.

oneAPI.jl, at some point, will include fixes for the logical indexing problem, as there has been movement with using KernelAbstractions.jl as the backend for GPUArrays.jl.

KomaMRIBase/src/motion/motionlist/Motion.jl
```diff
@@ -172,5 +186,5 @@ function Base.view(m::Motion, p)
 end
 
 # Auxiliary functions
-times(m::Motion) = times(m.time)
+times(m::Motion{T}, ::Type{T}) where {T<:Real} = times(m.time)
```
cncastillo, line 189: I am confused, why do you need this type T parameter?

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -1035,8 +1035,8 @@ function plot_phantom_map(
     max_time_samples=100,
     kwargs...,
 )
-    function interpolate_times(motion)
-        t = times(motion)
+    function interpolate_times(motion, T)
```
cncastillo, line 1038: Same here. You are modifying Base, Core, and Plots. This should be only a change in Base.

cncastillo (changes requested):
This PR has too many changes, based on the title I would have expected it to only change KomaMRIBase. Please separate the other changes, to register this new fix in base, without the rest.

KomaMRIPlots/src/ui/DisplayFunctions.jl
pvillacorta: This modification (obviating the T-parameter problem already solved in [cd8f751](https://github.com/JuliaHealth/KomaMRI.jl/pull/499/commits/cd8f75157b76181b41a3f8a061e9c1cf17cc94a2)) in `plot_phantom_map` came from [this comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/495#issuecomment-2413961563) on the phantom documentation PR. I thought it won't be a bad idea to solve it in this PR, even changing KomaMRIPlots.

Regarding de changes in core, the only change is the removal of the adapt function for NoMotion in Functors.jl, since it is no longer necessary

KomaMRIBase/src/motion/motionlist/Motion.jl
pvillacorta: Anyway, I can leave the change of the plot to the other PR.

KomaMRIPlots/src/ui/DisplayFunctions.jl
pvillacorta: Anyway, I can leave the change of the plot to the other PR.

cncastillo:
I think this PR should be working fine, the oneAPI.jl problems can wait.

@pvillacorta Do you think it is ready to merge?

KomaMRIBase/src/motion/MotionList.jl
```diff
@@ -133,22 +138,22 @@ function get_spin_coords(
 end
 
 """
-    times = times(motion)
```
cncastillo, line 136: Docs are not correct.

cncastillo (changes requested):
I think this PR also changes the name of a folder, which confused the diif's. please maintain the the name of the folders and files.

Also verify if all the docs changes are correct.

pvillacorta:
I think it is ready now

cncastillo:
I think the plotting changes came back.

pvillacorta:
> I think the plotting changes came back (fixing another bug).

Oh, I didn't get to remove those changes at any point in the end. Don't you think it's a good idea to take the opportunity to fix those bugs as well?

cncastillo:
Not the best practice, but let's change the title of the PR and list the bugs this PR is solving. So at least someone looking at it can understand.

PR #516, opened 2024-12-05
Fix flow-related functions

pvillacorta:
This PR is intended to solve some problems with flow experiments. Essentially:

- [x] We need to correct how the interpolation of the reset flags is performed. The `get_mask` function was incorrectly implemented (there was an unnecessary `cumsum`, which in fact caused the reset flag to be 1 at less time points than it should).
- [x] We need to ensure that, during the discretization of the sequence, time points corresponding specifically to spin jumps are included. Otherwise, we might miss a jump if no evaluation time instant falls within the interval where the reset bit should equal 1.


KomaMRIBase/src/datatypes/simulation/DiscreteSequence.jl
```diff
@@ -87,7 +87,7 @@ based on simulation parameters.
 """
 function discretize(seq::Sequence; sampling_params=default_sampling_params(), add_times=[])
     t, Δt      = get_variable_times(seq; Δt=sampling_params["Δt"], Δt_rf=sampling_params["Δt_rf"])
-    sort!(unique!(append!(t, add_times)))
+    sort!(append!(t, add_times))
```
cncastillo, line 90: don't add key points here, it should be added in `get_variable_times`.

cncastillo (changes requested):

KomaMRIBase/src/motion/NoMotion.jl
```diff
@@ -49,3 +49,4 @@ function get_spin_coords(
 ) where {T<:Real}
     return x, y, z
 end
+get_jump_times(::NoMotion) = []
```
cncastillo, line 52: This is an array of type Any.

cncastillo (changes requested):

PR #518, opened 2024-12-07
Limited size of large assets in light mode

Stockless:
In light mode, some images got out of bounds. Fixed it by limiting it's maximum width in a @raw html block.


cncastillo:
And don't think this solves the core issue. Why is this happening only for light mode? Is there a problem with the CSS? Why is this problem not happening elsewhere?

cncastillo:
@Stockless has this been solved? I would try to approach it by using "inspect elements" and see which CSS is being overwritten in light mode.

cncastillo:
The solution is to change center-images.css to:

```css
img {
  display: block; /* Ensures the image behaves as a block element */
  margin-left: auto;
  margin-right: auto;
}
```

PR #520, opened 2024-12-09
Solve bug in `SpinRanges` when indexing a phantom with a BitVector 

pvillacorta:
This worked:
`obj.motion = Translate(5e-4, 6e-4, 7e-4, TimeRange(0.0, 1.0), SpinRange(obj.x .> 0))`

But this did not work, and it was not tested:
`obj = obj[obj.x .> 0]`

When executing this second line, the resulting phantom had lost its motion. Two additional method definitions for the `intersect_idx` function inside SpinSpan.jl can solve this. 

I also use this PR to solve a bug in `get_spin_coords(m::Motion)`. I initialized the displacement buffers as follows:
` ux, uy, uz = x .+ 0*t, y .+ 0*t, z .+ 0*t`
And this is not correct, since it unnecessarily adds the initial positions to the displacements.

This is how it is solved:
` ux, uy, uz = x .* (0*t), y .* (0*t), z .* (0*t)`


cncastillo:
I believe you can also use `similar(a, M, N)`.

 I will wait until the tests pass!

cncastillo:
I still see some AMDGPU tests failing

pvillacorta:
I think it's a random error that has nothing to do with the code.

cncastillo:
You are right, I re-run that job, to see if it is fixed.

Issue #409, opened 2024-06-14
An error of 'DivideError: integer division' occurs in the internal functions of simulate when using a sequence with spiral acquisition


jigna405:
### What happened?

We are simulating various Pulseq sequences through KomaMRI, all of which work perfectly except for the sequence with spiral acquisition. The error occurs when simulation starts, indicating an attempt to perform an integer division where the divisor is 0. The division causing the error is happening in the function rem(x::Int64, y::Int64) in the int.jl file. 

![Captura de pantalla 2024-06-14 114929](https://github.com/JuliaHealth/KomaMRI.jl/assets/172833138/0e2a2c51-33d2-4df0-b09a-daa6015da4d1)

The test sequence being used is spiral.seq from Pulseq without modification. The first simulation was conducted using an AMD Ryzen 5 5500U with Radeon Graphics 2.10 GHz. Then, to verify if the error was due to hardware incompatibility, a second simulation was performed using an Nvidia RTX 3060 TI, which presented the same error. 

### Environment

```Markdown
OS x86_64-w64-mingw32
Julia 1.10.4
KomaMRIPlots 0.8.0
KomaMRIFiles 0.8.2
KomaMRI 0.8.2
KomaMRICore 0.8.0
KomaMRIBase 0.8.4
```

cncastillo:
Hi @jigna405, thanks for reporting! I can confirm that I was able to reproduce this error. It is throwing an error when generating the raw-data (ISMRMRD.jl). As we technically do not know how to reconstruct an image just with the sequence, we need some additional information about the image for the recon (like the size Nx, Ny, Nz). If that information is not available we guess it. This is a case in which the estimation is incorrect:

```julia
julia> seq.DEF["Nx"]
12000

julia> seq.DEF["Ny"]
0  #<-----------The problem

julia> seq.DEF["Nz"]
2
```
You can avoid this guessing by specifying the dimension in the .seq. Specifically in the `[DEFINITIONS]` section. Try to specify 

```
...
[DEFINITIONS]
...
Nx 100
Ny 100
Nz 1
...
```

We will definitely need to fix this.

curtcorum:
It is interesting. The spiral.seq file used to work:

https://github.com/JuliaHealth/KomaMRI.jl/discussions/311#discussion-6251011

cncastillo:
Yeah I think that it is because I changed `KomaMRIFiles/src/Sequence/Pulseq.jl` from

```julia
[...]

PR #527, opened 2024-12-18
Limit 'NadcsPerImage' to be at least 1

Stockless:
As "NadcsPerImage" shouldn't ever be less than 1, it gets limited to be the at least 1. This implementation avoids a possible division by 0 in `ISMRMRD.jl`, Solving the Issue #409 :
```julia
if scan_counter % NadcsPerImage == 0 #For now only Nz is considered
    nz += 1 #another image
    scan_counter = 0 #reset counter
end
```


cncastillo:
Thanks!

PR #530, opened 2024-12-31
Remove Kaleido_jll compat entry

### LINKED ISSUE #386: Fix plotting on M1 MacOS (Kaleido_jll v0.1)
### What happened?

We currently use Kaleido_jll v0.1 so plotting works on Windows. This generates a problem with M1 MacOS.

### Environment

```Markdown
-
```

cncastillo:
This is meant to fix the issues with Apple Silicon Mac OS. Fixes #386.


Stockless:
I have tried this change locally in an Apple Silicon MacOS and it solved some issues to pass the KomaMRI tests, but the issues that are being solved at the PR #538 need to be solved to verify in GitHub Actions if this contribution is correct.

PR #531, opened 2025-01-02
Flexibilize the definition of motion time spans

pvillacorta:
This PR arises from the need to create motion patterns that are currently not replicable in Koma. An example is the figure below, taken from [this paper](https://jcmr-online.biomedcentral.com/articles/10.1186/1532-429X-16-48), which shows the temporal evolution of a contraction and torsion motion in the myocardium:

image

This time curve cannot be replicated either with the `TimeRange` structure or the `Periodic` structure, so a new structure called `TimeCurve` has been defined, which can replicate any arbitrary temporal behaviour.

In fact, the `TimeRange` and `Periodic` structures may no longer be necessary if the efficiency of the `unit_time` function with a `TimeCurve` is better or at least equal to the corresponding `unit_time` implementations with `TimeRange` and `Periodic`. Efficiency checks are still pending.

Additionally, I want to add a field to account for pseudo-periodic motion patterns such as arrhythmias.

https://juliahealth.org/KomaMRI.jl/previews/PR531/


KomaMRIBase/src/motion/Motion.jl
```diff
@@ -65,7 +65,7 @@ end
 julia> translate = Translate(0.01, 0.02, 0.03, TimeRange(0.0, 1.0), SpinRange(1:10))
 ```
 """
-function Translate(dx, dy, dz, time=TimeRange(zero(eltype(dx))), spins=AllSpins())
+function Translate(dx, dy, dz, time=TimeRange(t_start=-oneunit(eltype(dx)), t_end=zero(eltype(dx))), spins=AllSpins())
```
cncastillo, line 68: Why do you need the start to be negative???

pvillacorta: This is to avoid having to include, in the `unit_time` function, a specific check for the case where `t_start == t_end`. I had thought that, by default, the motion would start at a negative `t` value and be completed at `t = 0`. In short, I had considered making it so that defining two identical time instants would not be supported, as it doesn't make much physical sense, and furthermore, Interpolations.jl does not support two identical nodes since it cannot determine which one comes before the other.

cncastillo: 0 and eps?

pvillacorta: It may be a better option. I will change it.

cncastillo (approved):

PR #536, opened 2025-01-23
Fix motion bugs and add `colorscale` keyword in `plot_image`

pvillacorta:
This PR solves some motion-related bugs that appeared in previous PRs.
Additionally, I added a `colorscale` keyword for the `plot_image` function (Let me know if I should include this in another PR or take advantage and include it in this one).

### DIFF
--- a/KomaMRIBase/src/datatypes/sequence/Grad.jl
+++ b/KomaMRIBase/src/datatypes/sequence/Grad.jl
@@ -226,6 +226,13 @@ Base.zero(::Type{Grad}) = Grad(0.0, 0.0)
 # Rotation
 Base.zero(::Grad) = Grad(0.0, 0.0)
 *(α::Real, x::Grad) = Grad(α * x.A, x.T, x.rise, x.fall, x.delay)
+*(α::Matrix, x::Array{Grad}) = begin
+    y = deepcopy(x)
+    for (i, g) in enumerate(y)
+        g.A = (α*x.A)[i]
+    end
+    return y
+end
 +(x::Grad, y::Grad) = Grad(x.A .+ y.A, max(x.T, y.T), max(x.rise, y.rise), max(x.fall, y.fall), max(x.delay, y.delay)) #TODO: solve this in a better way (by "stacking" gradients) issue #487
 # Others
 *(x::Grad, α::Real) = Grad(α * x.A, x.T, x.rise, x.fall, x.delay)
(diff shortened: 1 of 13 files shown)


cncastillo:
I think it's fine, is it ready to merge?

pvillacorta:
I also took advantage of this PR to add a method definition for the multiplication of an `Array` of `Grad`s by a rotation `Matrix` (https://github.com/JuliaHealth/KomaMRI.jl/pull/536/commits/36ab6c6fe8ab6010e376d168e376ad998e39b0b9).
This was necessary because rotating an array of gradients (in x, y, and z) with different durations caused all durations to become the same:

```julia-repl
julia> G1=Grad(3e-3,1e-3,0)
Grad(3.0 mT, 1.0 ms)

julia> G2=Grad(1.5e-3,2e-3,0)
Grad(1.5 mT, 2.0 ms)

julia> G3=Grad(4.5e-3,0.5e-3,0)
Grad(4.5 mT, 0.5 ms)

julia> g = [G1; G2; G3;;]
3×1 Matrix{Grad}:
 Grad(3.0 mT, 1.0 ms)
 Grad(1.5 mT, 2.0 ms)
 Grad(4.5 mT, 0.5 ms)

julia> R = [0. 0. 1.; 0. -1. 0.; 1. 0. 0.]
3×3 Matrix{Float64}:
 0.0   0.0  1.0
 0.0  -1.0  0.0
 1.0   0.0  0.0

julia> R * g
3×1 Matrix{Grad}:
 Grad(4.5 mT, 2.0 ms)
 Grad(-1.5 mT, 2.0 ms)
 Grad(3.0 mT, 2.0 ms)
```
I do not know if this is the expected behaviour, but I can assume it is not:
```julia
*(x::Sequence, A::Matrix{Float64}) = Sequence(A*x.GR, x.RF, x.ADC, x.DUR, x.DEF) #TODO: change this, Rotation fo waveforms is broken
*(A::Matrix{Float64}, x::Sequence) = Sequence(A*x.GR, x.RF, x.ADC, x.DUR, x.DEF) #TODO: change this, Rotation fo waveforms is broken
``` 
Now only the gradient amplitudes rotate, while the durations keep the same (I am not sure if this is correct for all cases):
```julia-repl
julia> R * g
3×1 Matrix{Grad}:
 Grad(4.5 mT, 1.0 ms)
 Grad(-1.5 mT, 2.0 ms)
 Grad(3.0 mT, 0.5 ms)
```

Please let me know if this change is correct, or tell me how you would do it and whether you would include it in this PR as well.
Thank you

cncastillo:
I don't think the gradient rotation is correct. Do you really need it to be like that?

pvillacorta:
I have needed this when creating a gre sequence and wanting to acquire different planes with it. In the initial basic sequence, which acquires the XY plane slice, there is a block with gradients of different durations. When rotating the sequence to acquire, for example, the YZ plane, all the gradients of the block end up having the same duration. I think this is an important use case.

PR #537, opened 2025-01-23
Faster Bloch GPU

rkierulf:
To keep my GPU programming skills sharp while applying to jobs, I've been working on a faster version of the Bloch GPU code I wrote this past summer. The first thing I noticed while working on this was that my earlier profiling wasn't 100% correct - in Julia, GPU kernel launches are all asynchronous, but they appear to be synchronous since kernels execute in the same order as they are submitted. So I needed to also add `KA.synchronize(backend)` before each `@signpost_event` to see which parts were taking the longest. The main parts taking the longest were these lines:

`Δϕ .= (Bz[:,2:end] .+ Bz[:,1:end-1]) .* Δt .* T(-π .* γ)`

`pre.Mxy[:,2:end] .= M.xy .* exp.(-seq_block.tp_ADC' ./ p.T2) .* cis.(ϕ_ADC)`

`pre.φ .= T(-π .* γ) .* (pre.B[:,1:end-1] .* seq.Δt')`

I think with the logical indexing of the Bz, B, and M.xy matrices, it is too hard for the compiler to generate efficient kernels for these operations that don't require allocating new memory. So writing simple kernels for these operations will be faster.

The other major thing I noticed was that the previous approach of trying to distribute work to one thread per time point **_and_** spin index is likely not optimal, since the previous implementation was memory bound. Each time we are creating a new matrix like in the line below:

`pre.ΔT1 .= exp.(-seq.Δt' ./ p.T1)` 

this requires `number of spins x number of time points` writes to global GPU memory, plus the same number of reads later. In theory, only `number of spins + number of time points` global memory reads are needed. Consider the following cumsum example:

```{julia}
using Metal, BenchmarkTools, KernelAbstractions

A = MtlArray(rand(Float32, (10000, 1000)))
B = similar(A)

@kernel function naive_cumsum!(B, @Const(A))
     i = @index(Global)
     T = eltype(A)

    cur_val = zero(T)
    for k ∈ 1:size(A, 2)
        @inbounds cur_val += A[i, k]
        @inbounds B[i, k] = cur_val
    end
end

@benchmark Metal.@sync cumsum!(B, A; dims=2)

BenchmarkTools.Trial: 402 samples with 1 evaluation per sample.
 Range (min … max):  11.520 ms …  21.151 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     12.301 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   12.433 ms ± 688.393 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                ▁▃ ▄█▄▇▂▆▄         ▁  ▁     ▅▁                  
  ▃▃▆▄▁▁▁▁▃▁▁▃▄▃██▆████████▇▅▅▄▆▅▄▆█████▆▆▅▇███▅▃▃▁▃▃▃▁▁▃▃▁▃▁▃ ▄
  11.5 ms         Histogram: frequency by time         13.4 ms <

 Memory estimate: 7.33 KiB, allocs estimate: 309.

@benchmark Metal.@sync naive_cumsum!(MetalBackend())(B, A, ndrange=size(A,1))

BenchmarkTools.Trial: 1836 samples with 1 evaluation per sample.
 Range (min … max):  1.690 ms …   3.963 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.864 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.713 ms ± 547.521 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                    ▇█                         
  ▂▂▄▂▃▃▂▃▄▅▅▅▄▄▇▅▃▃▃▂▃▃▂▂▂▂▂▃▄▄▄▃▄▄██▆▄▄▃▂▃▂▃▃▁▂▂▂▂▃▃▂▂▂▂▃▅▃ ▂
  1.69 ms         Histogram: frequency by time        3.76 ms <

 Memory estimate: 5.16 KiB, allocs estimate: 216.
```

The "naive" version outperforms the default Metal accumulate function which does a tree-based reduction even though it uses far less threads, since both kernels are memory bound and the naive version has a more efficient memory access pattern for this example.

With this in mind, I've rewritten the `run_spin_precession!` and `run_spin_excitation!` functions with a focus on minimizing the total number of global memory reads, so that in each block only the phantom and sequence arrays are read from and temporary values calculated from these arrays are stored in registers, not GPU global memory. For the signal output, we do still need to use a matrix of size `number of spins x number of time points` in global memory, but this can be reduced by a factor of the block size by doing a block-level reduction in shared memory before writing to the output. 

Testing the changes on my Mac M1, the benchmarks are about 3-5x faster, but they seem to make the most difference with a large phantom object - for example, if I get rid of the [1:10,000] part setting up the slice selection benchmark so it is using the full 3D brain phantom, then it is over 10x faster. There is also a strange issue where two of the tests are failing but only in my test environment - if I copy paste the test body into its own script then it produces the expected results. I haven't been able to figure out the reason for this yet.


cncastillo:
Wow amazing! Thanks @rkierulf ! Currently the CI is a little messed up, so I cannot merge it straight away. I benchmarked the previous implementation, with this code:

[GPU_Benchmarks_ISMRM2025.zip](https://github.com/user-attachments/files/18536991/GPU_Benchmarks_ISMRM2025.zip)

And got:
image

The previous method didn't scale very well. Do you think the changes you have made can improve this? There are clearly two regimes, flat and linear growth, not sure if they correspond to the memory- and compute-bound regimes. What are your thoughts on this?

rkierulf:
Interesting! My guess is that the transition from flat to linear growth begins when the simulation becomes large enough to fully saturate the resources of the GPU. After that, due to the memory bound nature of the problem the growth in time should be proportional to the number of additional reads + writes to GPU global memory per added spin index. Since the new implementation has far fewer GPU memory operations per spin index, I would expect the slope to be less steep. And hopefully, it is more efficient overall so the curve would shift somewhat to the right as well. I'd be curious to see the results of this script run again with the new changes.

Regarding the CI Failures:

- OneAPI and non-Buildkite CI failures: appear to be preexisting?
- Metal failures: these are the same two tests I was referencing that were only failing in my test environment, except here they are failing on Julia 1.10. Very strange, I'll try to investigate some more
- CUDA and AMD failures: both are failing on 1.10 while calling `KA.synchronize()` in `run_sim_time_iter!`. Maybe I need to be more conservative and call `KA.synchronize()` immediately after the kernels finish, I'll see if this fixes the issue

cncastillo:
Thanks for investigating. I will for rerun the benchmarks using this PR  when I have more time (moving to Stanford tomorrow, and then I have a conference in Washington DC).

I believe the problem with Julia 1.11 and oneAPI is: https://github.com/JuliaGPU/oneAPI.jl/issues/473. 

Recently they changed the internals of all the GPU packages to be based on KA.jl. onAPI.jl v2.0 should include these changes, so bumping the compat to include `2` might fix than error, but now sure. 

Don't mind the GitHub CI errors, @Stockless is currently fixing those.

rkierulf:
The tests are all passing now, except for the oneAPI tests with Julia 1.11. For the 1.11 tests overall, I'm not sure it is using the local versions of the packages. With the 1.10 tests I can see in the Manifest info printed before the tests are run:

```
[d0bc0b20] KomaMRIBase v0.9.0 `/var/lib/buildkite-agent/builds/gpuci-7/julialang/komamri-dot-jl/KomaMRIBase`
--
[4baa4f4d] KomaMRICore v0.9.0 `/var/lib/buildkite-agent/builds/gpuci-7/julialang/komamri-dot-jl/KomaMRICore`
```

With 1.11 it is:

```
[d0bc0b20] KomaMRIBase v0.9.0
--
[4baa4f4d] KomaMRICore v0.9.0
```

This may be a difference in how `Pkg.develop()` works in Julia 1.11 vs 1.10. I think the pipeline step to develop the packages needs to be done in the KomaMRICore test environment, by adding `--project=KomaMRICore/test` in the code below:

```
julia -e 'println("--- :julia: Instantiating project")
     using Pkg
     Pkg.develop([
         PackageSpec(path=pwd(), subdir="KomaMRIBase"),
         PackageSpec(path=pwd(), subdir="KomaMRICore"),
     ])'
```

cncastillo:
Hi Ryan! @Stockless  run this PR against master. It seems that it improves the results significantly! 

master (Left) ------------------ This PR (right) 
<p align="center">
  
  
</p>

We are currently re-running these with more spins so the differences are more apparent, but thanks for these nice improvements! :smile: 

Do you think you would be able to write a paragraph explaining in simple terms the key aspects that enabled these improvements (compared to array-based)? 

I think it would be really interesting to write a scientific paper with you as co-author. The public of the paper would be MRI experts, which are not really familiar with GPU programming (basically for them writing something in PyTorch is the same as GPU programming). So that paragraph could be used for both the docs, and also the paper.

rkierulf:
Great! I will add some more info to the docs when I have more time. In the meantime, it would be a good idea for someone with a Nvidia GPU to determine the optimal group size for the precession and excitation kernels by changing the simulation parameters "gpu_groupsize_precession" and "gpu_groupsize_precession". For now, I hardcoded the default group sizes based on what I found to be optimal (32 and 256 for precession / excitation), but if it varies based on the GPU backend then it would be a good idea to change to using a function which dispatches on the backend. As a reminder, the group size should be a multiple of 32 between 32 and 1024, and likely also a power of 2, so I would try 32, 64, 128, 256, 512, and 1024. [Profilng the kernel calls with NSight Compute](https://github.com/JuliaGPU/CUDA.jl/blob/master/docs/src/development/profiling.md) would also be helpful to shed more light on the nature of the kernels (register usage, memory throughput, etc.) and see if anything could be done to make the kernels faster, as well as provide more details that could go in a paper.

cncastillo:
Hi Ryan! I was trying this branch on our A100's at Stanford, and for some reason in some GPUs you may need to put some additional `@uniform`'s (issue: https://github.com/JuliaGPU/KernelAbstractions.jl/issues/575). Now, I am getting this error:

```julia
ERROR: a undefined variable error was thrown during kernel execution on thread (129, 1, 1) in block (58, 1, 1).
Stacktrace:
 [1] macro expansion at /home/cacp/.julia/packages/KernelAbstractions/sWSE0/src/KernelAbstractions.jl:242
 [2] gpu__reduce_nd_by_block! at /home/cacp/.julia/packages/KernelAbstractions/sWSE0/src/macros.jl:318
 [3] gpu__reduce_nd_by_block! at ./none:0
...
ERROR: LoadError: KernelException: exception thrown during kernel execution on device NVIDIA A100 80GB PCIe
```

Do you have any idea where they could be missing? 

Complete error message:
<details>
```julia
ERROR: LoadError: KernelException: exception thrown during kernel execution on device NVIDIA A100 80GB PCIe
Stacktrace:
  [1] check_exceptions()
    @ CUDA ~/.julia/packages/CUDA/1kIOw/src/compiler/exceptions.jl:39
  [2] device_synchronize(; blocking::Bool, spin::Bool)
    @ CUDA ~/.julia/packages/CUDA/1kIOw/lib/cudadrv/synchronization.jl:191
  [3] device_synchronize
    @ ~/.julia/packages/CUDA/1kIOw/lib/cudadrv/synchronization.jl:178 [inlined]
  [4] checked_cuModuleLoadDataEx(_module::Base.RefValue{Ptr{CUDA.CUmod_st}}, image::Ptr{UInt8}, numOptions::Int64, options::Vector{CUDA.CUjit_option_enum}, optionValues::Vector{Ptr{Nothing}})
    @ CUDA ~/.julia/packages/CUDA/1kIOw/lib/cudadrv/module.jl:18
  [5] CuModule(data::Vector{UInt8}, options::Dict{CUDA.CUjit_option_enum, Any})
    @ CUDA ~/.julia/packages/CUDA/1kIOw/lib/cudadrv/module.jl:60
  [6] CuModule
    @ ~/.julia/packages/CUDA/1kIOw/lib/cudadrv/module.jl:49 [inlined]
  [7] link(job::GPUCompiler.CompilerJob, compiled::@NamedTuple{image::Vector{UInt8}, entry::String})
    @ CUDA ~/.julia/packages/CUDA/1kIOw/src/compiler/compilation.jl:409
  [8] actual_compilation(cache::Dict{…}, src::Core.MethodInstance, world::UInt64, cfg::GPUCompiler.CompilerConfig{…}, compiler::typeof(CUDA.compile), linker::typeof(CUDA.link))
    @ GPUCompiler ~/.julia/packages/GPUCompiler/OGnEB/src/execution.jl:262
  [9] cached_compilation(cache::Dict{Any, CuFunction}, src::Core.MethodInstance, cfg::GPUCompiler.CompilerConfig{GPUCompiler.PTXCompilerTarget, CUDA.CUDACompilerParams}, compiler::Function, linker::Function)
    @ GPUCompiler ~/.julia/packages/GPUCompiler/OGnEB/src/execution.jl:151
 [10] macro expansion
    @ ~/.julia/packages/CUDA/1kIOw/src/compiler/execution.jl:380 [inlined]
 [11] macro expansion
    @ ./lock.jl:273 [inlined]
 [12] cufunction(f::GPUArrays.var"#gpu_broadcast_kernel_cartesian#41", tt::Type{Tuple{KernelAbstractions.CompilerMetadata{…}, SubArray{…}, Base.Broadcast.Broadcasted{…}}}; kwargs::@Kwargs{always_inline::Bool, maxthreads::Nothing})
    @ CUDA ~/.julia/packages/CUDA/1kIOw/src/compiler/execution.jl:375
 [13] macro expansion
    @ ~/.julia/packages/CUDA/1kIOw/src/compiler/execution.jl:112 [inlined]
 [14] (::KernelAbstractions.Kernel{…})(::SubArray{…}, ::Vararg{…}; ndrange::Tuple{…}, workgroupsize::Nothing)
    @ CUDA.CUDAKernels ~/.julia/packages/CUDA/1kIOw/src/CUDAKernels.jl:103
 [15] _copyto!
    @ ~/.julia/packages/GPUArrays/uiVyU/src/host/broadcast.jl:71 [inlined]
 [16] materialize!
    @ ~/.julia/packages/GPUArrays/uiVyU/src/host/broadcast.jl:38 [inlined]
 [17] materialize!(dest::SubArray{ComplexF32, 2, CuArray{…}, Tuple{…}, false}, bc::Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{…}, Nothing, typeof(identity), Tuple{…}})
    @ Base.Broadcast ./broadcast.jl:880
 [18] run_spin_precession!(p::Phantom{…}, seq::DiscreteSequence{…}, sig::SubArray{…}, M::Mag{…}, sim_method::Bloch, groupsize::Int64, backend::CUDABackend, pre::KomaMRICore.BlochGPUPrealloc{…})
    @ KomaMRICore ~/.julia/packages/KomaMRICore/KOY1Y/src/simulation/SimMethods/Bloch/gpu/BlochGPU.jl:44
 [19] (::KomaMRICore.var"#33#35"{Phantom{…}, DiscreteSequence{…}, SubArray{…}, Mag{…}, Bloch, Int64, CUDABackend, KomaMRICore.BlochGPUPrealloc{…}, Vector{…}})(::Tuple{Int64, UnitRange{…}})
    @ KomaMRICore ~/.julia/packages/KomaMRICore/KOY1Y/src/simulation/SimulatorCore.jl:96
 [20] next
    @ ~/.julia/packages/Transducers/fnznF/src/library.jl:54 [inlined]
 [21] next
    @ ~/.julia/packages/Transducers/fnznF/src/core.jl:775 [inlined]
 [22] macro expansion
    @ ~/.julia/packages/Transducers/fnznF/src/core.jl:181 [inlined]
 [23] __foldl__
    @ ~/.julia/packages/Transducers/fnznF/src/processes.jl:157 [inlined]
 [24] foldl_basecase
    @ ~/.julia/packages/Transducers/fnznF/src/processes.jl:361 [inlined]
 [25] _reduce_basecase(rf::Transducers.Reduction{Transducers.Map{…}, Transducers.BottomRF{…}}, init::Nothing, reducible::Transducers.SizedReducible{Base.Iterators.Enumerate{…}, Int64})
    @ Transducers ~/.julia/packages/Transducers/fnznF/src/threading_utils.jl:58
 [26] _reduce(ctx::Transducers.CancellableDACContext, rf::Transducers.Reduction{Transducers.Map{…}, Transducers.BottomRF{…}}, init::Nothing, reducible::Transducers.SizedReducible{Base.Iterators.Enumerate{…}, Int64})
    @ Transducers ~/.julia/packages/Transducers/fnznF/src/reduce.jl:139
 [27] _transduce_assoc_nocomplete
    @ ~/.julia/packages/Transducers/fnznF/src/reduce.jl:131 [inlined]
 [28] #transduce_assoc#253
    @ ~/.julia/packages/Transducers/fnznF/src/reduce.jl:108 [inlined]
 [29] transduce_assoc
    @ ~/.julia/packages/Transducers/fnznF/src/reduce.jl:84 [inlined]
 [30] foldxt
    @ ~/.julia/packages/Transducers/fnznF/src/reduce.jl:235 [inlined]
 [31] foreach(f::KomaMRICore.var"#33#35"{Phantom{…}, DiscreteSequence{…}, SubArray{…}, Mag{…}, Bloch, Int64, CUDABackend, KomaMRICore.BlochGPUPrealloc{…}, Vector{…}}, itr::Base.Iterators.Enumerate{Vector{…}}; kw::@Kwargs{})
    @ ThreadsX.Implementations ~/.julia/packages/ThreadsX/Bml38/src/foreach.jl:118
 [32] foreach(f::KomaMRICore.var"#33#35"{Phantom{…}, DiscreteSequence{…}, SubArray{…}, Mag{…}, Bloch, Int64, CUDABackend, KomaMRICore.BlochGPUPrealloc{…}, Vector{…}}, itr::Base.Iterators.Enumerate{Vector{…}})
    @ ThreadsX.Implementations ~/.julia/packages/ThreadsX/Bml38/src/foreach.jl:118
 [33] run_spin_precession_parallel!(obj::Phantom{…}, seq::DiscreteSequence{…}, sig::SubArray{…}, Xt::Mag{…}, sim_method::Bloch, groupsize::Int64, backend::CUDABackend, prealloc::KomaMRICore.BlochGPUPrealloc{…}; Nthreads::Int64)
    @ KomaMRICore ~/.julia/packages/KomaMRICore/KOY1Y/src/simulation/SimulatorCore.jl:95
 [34] run_sim_time_iter!(obj::Phantom{…}, seq::DiscreteSequence{…}, sig::CuArray{…}, Xt::Mag{…}, sim_method::Bloch, backend::CUDABackend; Nblocks::Int64, Nthreads::Int64, precession_groupsize::Int64, excitation_groupsize::Int64, parts::Vector{…}, excitation_bool::Vector{…}, w::Nothing)
    @ KomaMRICore ~/.julia/packages/KomaMRICore/KOY1Y/src/simulation/SimulatorCore.jl:206
 [35] macro expansion
    @ ./timing.jl:581 [inlined]
 [36] simulate(obj::Phantom{Float64}, seq::Sequence, sys::Scanner; sim_params::Dict{String, Any}, w::Nothing)
    @ KomaMRICore ~/.julia/packages/KomaMRICore/KOY1Y/src/simulation/SimulatorCore.jl:394
 [37] simulate
    @ ~/.julia/packages/KomaMRICore/KOY1Y/src/simulation/SimulatorCore.jl:335 [inlined]
 [38] macro expansion
    @ ~/.julia/packages/Suppressor/cKnGx/src/Suppressor.jl:28 [inlined]
 [39] var"##core#232"(obj#231::Phantom{Float64})
    @ Main ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:598
 [40] var"##sample#233"(::Tuple{Phantom{Float64}}, __params::BenchmarkTools.Parameters)
    @ Main ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:607
 [41] _lineartrial(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters; maxevals::Int64, kwargs::@Kwargs{})
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:186
 [42] _lineartrial(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters)
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:181
 [43] #invokelatest#2
    @ ./essentials.jl:1055 [inlined]
 [44] invokelatest
    @ ./essentials.jl:1052 [inlined]
 [45] #lineartrial#46
    @ ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:51 [inlined]
 [46] lineartrial
    @ ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:50 [inlined]
 [47] tune!(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters; progressid::Nothing, nleaves::Float64, ndone::Float64, verbose::Bool, pad::String, kwargs::@Kwargs{})
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:299
 [48] tune!
    @ ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:288 [inlined]
 [49] tune!(b::BenchmarkTools.Benchmark)
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:288
 [50] top-level scope
    @ ~/.julia/packages/BenchmarkTools/1i1mY/src/execution.jl:461
 [51] include(fname::String)
    @ Main ./sysimg.jl:38
in expression starting at /home/cacp/KomaBenchmarks/test.jl:37
Some type information was truncated. Use `show(err)` to see complete types.
```
</details>

rkierulf:
For now, I would recommend testing with KernelAbstractions v0.9.33 - the 9.34 change made a big difference in performance, since it rearranges the way valid index checks work to ensure that `@synchronize` is reached by all threads in a threadgroup. However, these added checks can cause a lot of additional branching, especially with code such as this:

```{julia}
if N >= 512u16
    if i <= 256u16
        s_data1[i] = s_data1[i] + s_data1[i + 256u16]
        s_data2[i] = s_data2[i] + s_data2[i + 256u16]
    end
    @synchronize()
end
if N >= 256u16
    if i <= 128u16
        s_data1[i] = s_data1[i] + s_data1[i + 128u16]
        s_data2[i] = s_data2[i] + s_data2[i + 128u16]
    end
    @synchronize()
end
...
```

For the precession kernel in this PR which uses `@synchronize`,  I am already launching the kernel with slightly more threads than needed to ensure the number of threads is a multiple of the groupsize, and doing my own valid index checking (the `i <= N_SPINS` part) for the actual computation. The `@synchronize` statements are outside of `i <= N_SPINS` so they will be reached by all threads, so the added checks in KA 0.9.34 are unnecessary and can be disabled by using `unsafe_indices=true` to recover the lost performance. I have a commit ready to go which does this and I think should also fix the issue you're seeing. However, this PR also added AcceleratedKernels.jl as a dependency, and their reduce function is also a lot slower with the 0.9.34 changes. So eventually we will either need them to adapt to the 0.9.34 changes by using `unsafe_indices`, or switch to using the base version of reduce which is slightly slower.

cncastillo:
Pinning KA.jl worked! 

I run tests with 100 million spins! At 10 million spins we start getting "Out of Memory" errors with array-based, which shows the power of kernel-based methods.

gpu_benchmarks

gpu_speedup

Kernel-based is around 50x faster than array-based and 100x faster than CPU multi-threading. Great results! :partying_face: We can simulate at least 90 million spins more using kernel-based compared to array-based.

rkierulf:
Ok, this should now be ready for review. I added some info to the docs, and also changed the default groupsizes for both kernels to 256 to keep occupancy high. I believe the kernels are more compute-bound at this point, so it is hard to say what could be done to speed them up further without profiling with NSight compute.

AMDGPU has also started failing in the most recent builds, with different errors for 1.10 and 1.11. The 1.10 error is occurring in the 1D cumsum in BlochDict and looks like a bug in their implementation, so I've written this issue: https://github.com/JuliaGPU/AMDGPU.jl/issues/742. The 1.11 issue is happening in the old BlochGPU code (as I mentioned in my earlier comment, it is not using the local version of the package) which is being replaced in this PR, so it doesn't seem like too much of a concern.

pvillacorta:
Hi! Yes, I added that comment when implementing spin resetting for flow simulations, but it is not implemented inside the kernel-based function yet. So that comment should be kept :)

cncastillo:
I think this is ready to merge. 

@rkierulf anything that we could have missed? do you also think this PR is ready to be merged?

Cheers!

rkierulf:
Yep, this should be ready to merge!

PR #538, opened 2025-01-27
Fix CI tests for macoOS and ubuntu OS.

Stockless:
Fixed the CI test for macOS replacing the OS version from macOS-latest to macOS-13, and the test for Ubuntu adding a macro to avoid the generation of a sandbox in the creation of the gui window, that is causing an issue due to a bug in the latest version of Ubuntu (Ubuntu 24.04.1)


.github/workflows/CI.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         version:
           - '1.10' # Replace this with the minimum Julia version that your package supports. E.g. if your package requires Julia 1.5 or higher, change this to '1.5'.
           - '1' # Leave this line unchanged. '1' will automatically expand to the latest stable 1.x release of Julia.
-        os: [ubuntu-latest, windows-latest, macos-latest]
```
cncastillo, line 22: Why Is macos-latest not working? Is it because of Kaileido_jll? Check the PR I created about that.

test/runtests.jl
```diff
@@ -38,10 +38,10 @@ end
 @testitem "KomaUI" tags=[:koma] begin
 
     using Blink
+    # Opens KomaUI
 
-    # Opens UI
-    w = KomaUI(return_window=true) 
-
+    @unsafe_blink
```
cncastillo, line 43: Yo actually fix the bug (opposed to only fix the tests) the call to the macro needs to be inside the KomaUI function.

cncastillo:
@Stockless can you reply to the comments?

cncastillo:
please don't include changes from other PRs without merging them properly, so the ownership of the changes is not changed.

cncastillo:
Hi @Stockless, What is the state of this?

cncastillo:
This branch is already outdated. Please don't add new dependencies unless strictly necessary.

KomaMRIPlots/Project.toml
```diff
@@ -20,7 +20,7 @@ KomaPlotsPlutoPlotlyExt = "PlutoPlotly"
 
 [compat]
 Interpolations = "0.13, 0.14, 0.15"
-Kaleido_jll = "0.1"
```
cncastillo, line 23: Why are you putting this here? The solution should be to remove the compat. This is already done in #530

Project.toml
```diff
@@ -14,6 +14,7 @@ KomaMRIPlots = "76db0263-63f3-4d26-bb9a-5dba378db904"
 MAT = "23992714-dd62-5051-b70f-ba57cb901cac"
 MRIReco = "bdf86e05-2d2b-5731-a332-f3fe1f9e047f"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
```
cncastillo, line 17: Why are you adding this?

src/KomaMRI.jl
```diff
@@ -8,7 +8,9 @@ using Reexport
 import KomaMRICore: update_blink_window_progress!
 
 # GUI
-using Blink, Interact, AssetRegistry
+using Blink, Interact, AssetRegistry, Blink.AtomShell
+using Sockets: @ip_str
+using Blink.AtomShell: port, inspector, try_connect, electron, initcbs, mainjs
```
cncastillo, line 13:
```suggestion
using Blink, Interact, AssetRegistry
```

src/ui/UnsafeBlink.jl
```diff
@@ -0,0 +1,34 @@
+module MacroUnsafeBlink
```
cncastillo, line 1: Why is this a module?

KomaMRIPlots/Project.toml
Stockless: I did remove Kaleido_jll to test if that was the reason why the CI test in macOS-latest wasn't running, but when you told me that this change shouldn't be included because it is from another PR, I did an stepback in this commit [c0ce80e](https://github.com/JuliaHealth/KomaMRI.jl/pull/538/commits/c0ce80e66c023d4d2ab6b0ec6a32db5fe874c316)

cncastillo: Yes, but now it says `Kaleido_jll = "0.2.1"` that is not the same of `Kaleido_jll = "0.1"`

Project.toml
Stockless: I'm adding Sockets because it is part of the implementation of a macro to disable the sandbox in the window generation, creating a workaround for the connection bug as it is described in this [closed issue](https://github.com/Eben60/PackageMaker.jl/issues/1)

cncastillo: Sockets should already be included in Blink. Use `using Blink.Sockets`

src/ui/UnsafeBlink.jl
Stockless: I'm currently working in the implementation of the macro that fixed CI test in commit [c5117f0](https://github.com/JuliaHealth/KomaMRI.jl/pull/538/commits/c5117f0a76b7270e2dc19eb14aad0dcdd5647f97) inside the KomaUI function, and it's easier for me to run the different test from a module and import it. In a later version it should be incorporated directly in the function instead of the module import.

cncastillo: OK, please read carefully this https://jkrumbiegel.com/pages/2021-06-07-macros-for-beginners/ to understand better how macros work.

.github/workflows/CI.yml
Stockless: I tested removing the package in [e208255](https://github.com/JuliaHealth/KomaMRI.jl/pull/538/commits/e208255f50f651147a23e8109ce1177efcdc4a98) , but that wasn't the reason. I still don't know why it doesn't work in the latest version. I'll keep it in macOS-13 until it's solved.

KomaMRIPlots/Project.toml
cncastillo: Be careful is `Kaleido_jll = "0.1"` not `Kaleido_jll = "0.1.0"`

cncastillo:
Please reply to comments @Stockless.

Stockless:
Currently, the CI test are passing in [a16cffa](https://github.com/JuliaHealth/KomaMRI.jl/pull/538/commits/a16cffa9b9080df494f93221da651e92ac1a35f8) , but it's using macOS-13 instead of macOS-latest.
I'm working on fixing it to have the test with the latest version in every tested OS.

cncastillo:
@Stockless I was requested to review but the comments haven't been answered. What is the state of this?

cncastillo:
I will close this, to prefer #565.

PR #558, opened 2025-04-02
WIP : Implement labels

aTrotier:


aTrotier:
@cncastillo can you take a look at the initial implementation before going further.

image

aTrotier:
@cncastillo forget to push the rest !

cncastillo:
Sure! Thanks for implementing this, I will take a look later today.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -390,12 +434,32 @@ function read_seq(filename)
...
+                        labelsetLibrary = read_labels(io;eventLibrary=labelsetLibrary)
+                    elseif startswith(extension, "LABELINC")
+                        id = parse(Int, extension[9:end])
+                        extensionType[id] = Dict("data"=>"LABELINC")
+                        labelincLibrary = read_labels(io; eventLibrary=labelincLibrary)
+                    elseif startswith(extension, "DELAYS")
+                        @warn "DELAYS extension is not handle"
+                    else
+                        @warn "Ignoring unknown extension, input string: $extension"
+                    end
```
cncastillo, line 459: It would be nice to do this with multiple dispatch, as extensions are meant to be extended. So if the user creates a new extension called "MyEXT" they can just define maybe `read_extenstion(..., Val("MyEXT"))` or other functions like that, so they don't need to modify the code.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -706,10 +775,47 @@ function get_block(obj, i)
...
+        else
+            @warn "Extension type not implemented"
+        end
+
+        if next_id == 0
+            break
+        else
+            type, ref, next_id = extensionLibrary[next_id]["data"]
+        end
+    end
```
cncastillo, line 818: It would be nice to do this with multiple dispatch, as extensions are meant to be extended. So if the user creates a new extension called "MyEXT" they can just define maybe `read_extenstion(..., Val("MyEXT"))` or other functions like that, so they don't need to modify the code.

cncastillo (commented):
Hi! This looks awesome. I just have a quick nitpick, I added comments to some code blocks that could be benefited by the use of multiple dispatch. I don't have a clear idea yet, so it is only a comment. But feel free to continue developing this.

aTrotier:
Labels are now implemented
image

The label field is first empty and filled on request (if not empty).

aTrotier:
I am able to use the label for multi-slice reconstruction with KomaMRI but we are limited by something in MRIReco.jl :
The multi-slice reconstruction from MRIReco expects that all slices has the same trajectory (when I want to have different trajectory I manually label the raw.profiles along the rep or contrast dimensions which handles different trajectory).

In my local example the trajectory is the same so I just "cut" the trajectory at 1/3 of the total traj
```julia
acq = AcquisitionData(raw)
acq.traj[1].nodes = acq.traj[1].nodes[:,1:4096]

params = Dict{Symbol,Any}()
params[:reco] = "direct"
params[:encodedSize] = acq.encodingSize[1:2]
params[:densityWeighting] = true

im_reco = reconstruction(acq,params)
```

image

aTrotier:
I removed the label from Sequence struct. It is easier to handle that way and is not directly connected to the sequence (issue with subset of sequence, or if the user already extract the label and continue to add block etc)

now the label can be extracted with : `get_label(seq)` and returns a Vector{AdcLabels} (and `maximum(label)`)

aTrotier:
I now estimates the center directly from ktraj it works for both cases with my simple example (I need a partial-fourier acquisition along the readout in order to see if it really works)

2 points remains to be added in the reconstruction part :
`raw2.params["trajectory"] = "cartesian"`  -> nothing to do here except adding that in the documentation

`raw2.params["encodedSize"] = [seq.DEF["Nx"],seq.DEF["Ny"]]` -> The Nx/Ny from seq.DEF (calculated during the conversion from Pulseq) are different to the one calculated here in ISMRMRD.jl :
```julia
    if haskey(seq.DEF, "FOV")
        FOVx, FOVy, _ = seq.DEF["FOV"] #[m]
        if FOVx > 1 FOVx *= 1e-3 end #mm to m, older versions of Pulseq saved FOV in mm
        if FOVy > 1 FOVy *= 1e-3 end #mm to m, older versions of Pulseq saved FOV in mm
        Nx = round(Int64, FOVx / Δx[1])
        Ny = round(Int64, FOVy / Δx[2])
    else
        FOVx = Nx * Δx[1]
        FOVy = Ny * Δx[2]
    end
```

64,64 vs 63,63 (from the FOV). Should I put a `ceil` instead of `round` ?  @cncastillo

cncastillo:
Oh sorry for the late reply. 

> 64,64 vs 63,63 (from the FOV). Should I put a ceil instead of round ? @cncastillo

I don't have a strong preference, `ceil` could make more sense.

So can this be used in the UI already? Or is there something that needs to be modified for the reconstruction to incorporate the labels?

aTrotier:
My example with 3 slices with SLC / LIN (/ PAR) set correctly can be reconstructed by KomaUI() with `obj_ui[] = brain_phantom3D()` and I can slide along the 3 slices in the UI with and without defining the LIN (SLC are still defined)

You can give it a try with the attached sequences :
[Archive.zip](https://github.com/user-attachments/files/19898671/Archive.zip)

I am pretty sure it will failed at some point if the SLC label are not defined or if it is a 3D acquisition without label. 
We need to be careful and use more test sequences.

cncastillo:
Hi @aTrotier Thanks for the enhancements. 

> Should we add an example to show how to use the label for reconstruction ? Like removing ADC lines with labels -> ECO = 1 or something like that

I think a literate tutorial would be great. 

> It would be great if we could merge this feature before summer (I would like to start writing my MRI course with this possibility)

Awesome! I have no problems merging this before summer. @Stockless can you make sure this is merged asap and tag new versions so it is usable by summer?

JanWP:
Hello everyone,

Just stopping by to say a big thanks for this PR !
It's very timely: For the past few days I have been trying to implement an AFI sequence, which simply consists of two interleaved GRE acquisitions with different TRs. The k-space lines from the two images are interleaved, just like for multi-echo data, and I was struggling with the reconstruction.

With this PR, it worked perfectly. I have to use the ECO label for now (I think it's preferable to use SET, and that's what I tried initially, but it's not implemented on the reconstruction side yet). There was only one tiny issue: I had to change the "EXT.jl" to "Ext.jl" [here](https://github.com/aTrotier/KomaMRI.jl/blob/b8b212f0e5788495ac49becd194da4ccd0344958/KomaMRIBase/src/KomaMRIBase.jl#L25).

KomaMRIFiles/src/Sequence/Pulseq.jl
cncastillo: @Stockless can you create an issue including these to comments? "Make extensions extensible". So I don't forget!

KomaMRIFiles/src/Sequence/Pulseq.jl
cncastillo: @Stockless can you create an issue including these to comments? "Make extensions extensible". So I don't forget!

aTrotier:
@cncastillo Seems good on my side. 

Can you review the documentation ? :)

cncastillo:
That would be very cool actually! Do you want to use the drop-down to activate-deactivate text annotations showing the labels?

cncastillo:
Oh ok, that is perfect! It is more important for that information to be displayed somehow, and then we can change how it works.

aTrotier:
Last commit gives this visualization :
image

image

you can test it with this file : 
[gre_label_pypulseq_with_LIN.zip](https://github.com/user-attachments/files/21485629/gre_label_pypulseq_with_LIN.zip)

Can you give it a try ? @cncastillo

cncastillo:
Wow it is working very well! :rocket: 

I tried it using:
```julia-repl
pkg> add https://github.com/aTrotier/KomaMRI.jl:KomaMRIBase#extension_label
pkg> add https://github.com/aTrotier/KomaMRI.jl:KomaMRICore#extension_label
pkg> add https://github.com/aTrotier/KomaMRI.jl:KomaMRIFiles#extension_label
pkg> add https://github.com/aTrotier/KomaMRI.jl:KomaMRIPlots#extension_label
pkg> add KomaMRI
```

Comments below:

(1) Is there a way to show "Labels:" or something like that on top of the drop down?
image

(2) On  the UI, we use "darkmode", so the black "x"'s for the labels are not very visible. I would try to use one of these colors that change based on the mode:

```julia
bgcolor, text_color, plot_bgcolor, grid_color, sep_color = theme_chooser(darkmode)
```

image

(3) It seems that after clicking one of the labels in the drop down, some events that are generally hidden by default become visible (RF freq and RF phase). I would try to use the same `visible="legendonly"` for those.

image

(4) This one might be difficult, but it would be nice if the drop down only showed the Labels actually present in the sequence, and no dropdown if non are present. Alternatively, put in bold the ones that exist (show **LIN** instead of LIN). Whichever is easiest.

Thank you for your contribution!!

cncastillo:
1. Yeah this one doesn't look thaaat good, it moves if i resize the window, but we can leave it as it is for now.
2. cool!
3. It seems to be ok now, RF FM and AM are not plotted when selecting labels.
4. cool!

Ready to merge after the conflicts are solved :+1:.

aTrotier:
1. yeah sorry for that I tried to fix it but it looks like all the position need to be "fixed" and cannot be related to another object (here the button). I am really not familiar with PlotlyJS.

An idea to fix it : because we only show the activated label maybe we can put them directly as a legend ? In that case we should juste differentiate them with color / markershape ?

cncastillo:
1. Maybe changing option "none" to "Labels (none)", would be a good solution for now (?)

btw, tests seem to be failing (`ISMRMRD.jl:183`):

```julia
ERROR: LoadError: MethodError: no method matching -(::Nothing, ::Int64)
Closest candidates are:
  -(::BigFloat, ::Union{Int16, Int32, Int64, Int8})
   @ Base mpfr.jl:494
  -(::ChainRulesCore.NoTangent, ::Any)
   @ ChainRulesCore ~/.cache/julia-buildkite-plugin/depots/018dadfb-f4d5-466e-8c3f-55b5f9a75025/packages/ChainRulesCore/XAgYn/src/tangent_arithmetic.jl:61
  -(::Any, ::ChainRulesCore.NotImplemented)
   @ ChainRulesCore ~/.cache/julia-buildkite-plugin/depots/018dadfb-f4d5-466e-8c3f-55b5f9a75025/packages/ChainRulesCore/XAgYn/src/tangent_arithmetic.jl:50
  ...
Stacktrace:
  [1] signal_to_raw_data(signal::Array{ComplexF32, 3}, seq::KomaMRIBase.Sequence; phantom_name::String, sys::KomaMRIBase.Scanner, sim_params::Dict{String, Any}, ndims::Int64)
    @ KomaMRICore /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/komamri-dot-jl/KomaMRICore/src/rawdata/ISMRMRD.jl:183
  [2] simulate(obj::KomaMRIBase.Phantom{Float64}, seq::KomaMRIBase.Sequence, sys::KomaMRIBase.Scanner; sim_params::Dict{String, Any}, w::Nothing)
    @ KomaMRICore /var/lib/buildkite-agent/builds/amdgpu1-luraess-com/julialang/komamri-dot-jl/KomaMRICore/src/simulation/SimulatorCore.jl:434
```

aTrotier:
1) Looks better with your idea :)

Bug fixed

cncastillo:
Ok, the code coverage is complaining but it is a minor problem. I will run the other test, and if they pass we can merge this. 

@Stockless Can you tag new versions after this is merged?

cncastillo:
Ok there are some errors: 
```julia
`UndefKeywordError: keyword argument `label_to_show` not assigned
```

https://github.com/JuliaHealth/KomaMRI.jl/actions/runs/16718189812/job/47353477755?pr=558#step:18:376

```julia
UndefVarError: `get_label` not defined in `Main.var"##231"`
```
https://github.com/JuliaHealth/KomaMRI.jl/actions/runs/16718189812/job/47353477755?pr=558#step:12:175

```julia
 EXT: Test Failed at /home/runner/work/KomaMRI.jl/KomaMRI.jl/KomaMRIBase/test/runtests.jl:280
  Expression: SET_vec == vec([0 0 0 2 2 2])
   Evaluated: [0, 0, 0, 0, 0, 0] == [0, 0, 0, 2, 2, 2]
```

https://github.com/JuliaHealth/KomaMRI.jl/actions/runs/16718189812/job/47353477755?pr=558#step:12:175

KomaMRIFiles/test/runtests.jl
```diff
@@ -35,6 +35,8 @@ using TestItems, TestItemRunner, KomaMRIBase
         @test shape == shape2
 
         # Test label capability
+        using KomaMRIBase
```
aTrotier, line 38: I don't know why I have to add that here. KomaMRIBase is already imported no ?

cncastillo: Is `get_label` not defined if you don't? Maybe you forgot to export some function in KomaMRIBase.

aTrotier: even with the new export I still have to add the `using KomaMRIBase`. But I saw in the next @testitem that it is also the case : 

```julia
@testitem "Pulseq compat" tags=[:files, :pulseq] begin
    using MAT, KomaMRIBase, Suppressor
```
I guess it is something related with local scope :)

cncastillo: Yeah, I guess it is fine, specially if it is used elsewhere. I don't see any negative effects of this, so we can keep it like that.

KomaMRIBase/src/KomaMRIBase.jl
```diff
@@ -59,9 +60,9 @@ export get_spin_coords
 # Secondary
 export get_kspace, rotx, roty, rotz
 # Additionals
-export get_flip_angles, is_RF_on, is_GR_on, is_ADC_on
+export get_flip_angles, is_RF_on, is_GR_on, is_ADC_on, get_label
 # Sequence related
-export get_Mk, get_kspace, get_M0, get_M1, get_M2
+export get_Mk, get_kspace, get_M0, get_M1, get_M2, get_label
```
cncastillo: You are exporting `get_label` twice.

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -103,29 +112,29 @@ end
...
+Base.getindex(x::Sequence, i::Int) = Sequence(x.GR[:,i], x.RF[:,i], x.ADC[i], x.DUR[i],x.EXT[i], x.DEF)
+Base.getindex(x::Sequence, i::BitArray{1}) = any(i) ? Sequence(x.GR[:,i], x.RF[:,i], x.ADC[i], x.DUR[i], x.EXT[i],x.DEF) : nothing
+Base.getindex(x::Sequence, i::Array{Bool,1}) = any(i) ? Sequence(x.GR[:,i], x.RF[:,i], x.ADC[i], x.DUR[i], x.EXT[i],x.DEF) : nothing
 Base.lastindex(x::Sequence) = length(x.DUR)
 Base.copy(x::Sequence) where Sequence = Sequence([deepcopy(getfield(x, k)) for k ∈ fieldnames(Sequence)]...)
 
 #Arithmetic operations
 recursive_merge(x::Dict{K, V}) where {K, V} = merge(recursive_merge, x...)
 recursive_merge(x::Vector) = cat(x...; dims=1)
 recursive_merge(x...) = x[end]
```
cncastillo: Removing this `recursive_merge` functions could fix the code coverage. We are not using them anyway.

KomaMRICore/src/rawdata/ISMRMRD.jl
```diff
@@ -195,30 +207,26 @@ function signal_to_raw_data(
...
-            traj = ktraj[1:ndims, current:current+Nsamples-1]
-            #Acquired data, data::Array{Complex{Float32},2}, 1dim=numsamples, 2dim=coils
-            dat =  signal[current:current+Nsamples-1, :]
             #Saving profile
             push!(profiles, Profile(head, Float32.(traj), ComplexF32.(dat)))
             #Update counters
             scan_counter += 1
             current += Nsamples
-            if scan_counter % NadcsPerImage == 0 #For now only Nz is considered
+            if scan_counter % NadcsPerImage == 0 #For now only   is considered
```
cncastillo: Any reason to delete `Nz` here?

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -461,7 +528,7 @@ function read_seq(filename)
     fix_first_last_grads!(seq)
     # Final details
     # Hack for including extension and triggers, this will be done properly for #308 and #323
-    seq.DEF["additional_text"] = read_Extension(extensionLibrary, triggerLibrary) #Temporary hack
+    #seq.DEF["additional_text"] = read_Extension(extensionLibrary, triggerLibrary) #Temporary hack
```
cncastillo: Feel free to delete these lines.

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -31,8 +31,35 @@ function theme_chooser(darkmode)
 end
 
 function generate_seq_time_layout_config(
-    title, width, height, range, slider, show_seq_blocks, darkmode; T0
+    title, width, height, range, slider, show_seq_blocks, darkmode; T0,         label_to_show=0
```
cncastillo: Remove unnecessary spaces from `label_to_show`.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -706,10 +775,47 @@ function get_block(obj, i)
...
+
+    # Definitition
+    DEF = Dict{String,Any}()
 
     #Sequence block definition
-    s = Sequence(G,R,A,D,E)
+    s = Sequence(G,R,A,D,E,DEF)
     s
 end
+
```
cncastillo, line 719: Can you add a small doc string to this function, like in the other functions?

docs/src/explanation/5-seq-events.md
```diff
@@ -219,6 +219,103 @@ julia> plot_seq(seq; slider=false)
...
+# Display the sequence
+println(seq)
+```
+
+In this example, both `LabelInc` and `LabelSet` are added to the `EXTENSION` field of the sequence, allowing for more complex metadata management.
+
+By using `LabelInc` and `LabelSet`, you can effectively manage sequence metadata and ensure that your sequence is compatible with various reconstruction algorithms and formats.
+
+!!! warning
+    So far, **KomaMRI** EXTENSION only manage ADC labels. In future version, triggers and other specific Pulseq extension will be added.
```
cncastillo: Oh I thought that Triggers were also implemented. Were they too complicated to implement? I particularly would really like to use Triggers 😄. We could implement them later.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -606,6 +673,7 @@ function read_ADC(adcLibrary, i)
     A
 end
 
+#=
```
cncastillo: Feel free to remove this.

docs/src/explanation/5-seq-events.md
aTrotier: My bad, they are implemented :
```julia
mutable struct Trigger <: Extension 
  type::Int
  channel::Int
  d1::Float64
  d2::Float64
end
```

But not tested. We should add a test for that.

For a futur use I guess it is necessary to define add an additionnal timing block according to a temporal TTL signal.

docs/src/explanation/4-sequence.md
```diff
@@ -30,19 +30,20 @@ mutable struct Sequence
     RF::Array{RF,2}
     ADC::Array{ADC,1}
     DUR::Array{Any,1}
```
cncastillo: Wait I don't think `DUR::Array{Any,1}` is correct, it should be `Float64`.

cncastillo: Oh no, it is actually of type `Any` that is terrible, it should be `Float64`. That is a bug.

aTrotier:
> Ok, the PR is almost done 💪 thanks for all the work done, just a few minor comments.
> 
> Maybe this is already implemented, but can you make sure that the drop down is not displayed if the sequence has no labels? `if (isempty(labels))` or something like that.

done :)

cncastillo (changes requested):
Everything looks good, there is still an error in the docs. Could you investigate it?

https://github.com/JuliaHealth/KomaMRI.jl/pull/558/files#diff-1bb8f36e9d4c491fe4cfce00715868e3bbeb11286a57696de0bcec5e69f3e441

aTrotier:
release 1.5.1 of pulseq with new extension

> A minor release with some important improvements and bug fixes. New features:
> 
> Rotation extension
> RF shimming extension
> Plotting function shows triggers and digital output pulses
> Demo code for MR0 and Koma.jl signal simulations

I won't try to implement them for now. I really want to have that PR merged before the end of the year 😄

cncastillo:
I think the only remaining problem to merge this PR is that the generated figures on the literate example are too heavy:


```julia
┌ Error: Generated HTML over size_threshold limit: tutorial/07-label.md
│     Generated file size: 1.73 (MiB)
│     size_threshold_warn: 100.0 (KiB)
│     size_threshold:      200.0 (KiB)
│     HTML file:           tutorial/07-label/index.html
└ @ Documenter.HTMLWriter 
```

Could you plot a smaller section of the sequence? `plot_seq(seq[1:N])` (I am assuming that plot is generating the errors, haven't really checked)

Don't mind the CUDA error on Buildkite, that shoudl be fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/12007.

Stockless:
I fixed some issues with the reconstruction plots images, aligning them and removing the off-resonance. Also removed a semicolon that commented the `acqData.kdata |> size` line.

image→image

I think this PR is ready to be merged.

aTrotier:
Should we do something else before merging ?

Issue #550, opened 2025-03-24
Errors when generating documentation Pluto examples using Julia 1.11


Stockless:
The Pluto examples in `examples/4.reproducible_notebooks/` that are built at `docs/src/tutorial-pluto/` are presenting errors in the execution of its blocks.

cncastillo:
This is probably because the notebooks were generates using Julia 1.10 and the docs recently changed to be run using Julia 1.11.

One potential solution is to change the version of Julia used to generate the docs to Julia 1.10.

Stockless:
The issue comes from #499 , apparently creating a phantom object in Pluto still expects the notation previous to this PR:
```julia
obj = Phantom{Float64}(x=[])
```
instead of
```julia
obj = Phantom()
```
This bug is only inside the Pluto notebooks, not outside of it.

cncastillo:
This change is incorporated in Koma v0.9.1, the bug was introduced in 0.8.X, but the notebook should have the correct versions of the package. Like said previously, the problem is because is using a different Julia version and re-resolving the dependencies. Either:

- You run the documentation with Julia v1.10
- Upgrade the packages to KomaMRI v0.9.1

There is not need to change the code, as the code run with the package versions specified in the its Manifest.

pvillacorta:
Hello, I could try to spend more time trying to solve this, but for now, I can't find an explanation

pvillacorta:
Pluto notebooks usually have their own integrated Julia environment (Project.toml and Manifest.toml), which includes specific versions of each package used in them. I opened each of the three notebooks and updated the package versions following the steps in the "Updating packages" section of the Pluto documentation: https://plutojl.org/en/docs/packages/

By doing this, the packages are updated along with the environments corresponding to each notebook. They were using an old version of Koma (0.8), which is why it wasn’t working. This is what I did in #560.

However, updating PlutoPlotly to version 0.6 makes the `plot_phantom_map` function stop working (which is why I opened #559).

It seems that notebooks need to be updated as new versions of Koma are released.

Issue #559, opened 2025-04-02
`plot_phantom_map` not working with PlutoPlotly v0.6^


pvillacorta:
Since [this breaking change](https://github.com/JuliaPluto/PlutoPlotly.jl/pull/55) in PlutoPlotly, `plot_phantom_map` returns the following error when running on a Pluto notebook (thanks to @Stockless for discovering it):

![Image](https://github.com/user-attachments/assets/f717b629-2b5f-4d00-857a-fc41348e538f)

By the moment, until we find the solution (or post an issue in PlutoPlotly), Pluto notebooks using KomaMRIPlots should use PlutoPlotly@0.5 or lower.

disberd:
Hi all,

Sorry you had this issue for some months now. It should now be fixed in v0.6.4 of PlutoPlotly. Could you check if that is actually solving it?

Also please feel free to ping me directly if you have issues with PlutoPlotly, I try to at least provide some feedback relatively fast.

cncastillo:
Hi! Thank you so much for offering help, I will ping you if we have issues next time. CompatHelper has suggesting to update to 0.6 and we haven't done it yet because of this error, so we really didn't have a problem yet. Just now I had time to figure out what was going going on and put together a MWE :) 

I can confirm that this now works!

PR #560, opened 2025-04-02
Fix pluto notebooks

pvillacorta:
This PR fixes https://github.com/JuliaHealth/KomaMRI.jl/issues/550.

Package versions of the Pluto self-contained environments need to be updated with every Koma's new version release.

Also note that, while solving this, @Stockless discovered [this bug](https://github.com/JuliaHealth/KomaMRI.jl/issues/559), which is for now solved by limiting the maximum version of PlutoPlotly to v0.5.

### DIFF
--- a/examples/4.reproducible_notebooks/pluto-01-gradient-echo-spin-echo.jl
+++ b/examples/4.reproducible_notebooks/pluto-01-gradient-echo-spin-echo.jl
@@ -1,5 +1,5 @@
 ### A Pluto.jl notebook ###
-# v0.19.43
+# v0.20.5
 
 #> [frontmatter]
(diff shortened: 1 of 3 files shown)


Stockless:
This PR should include the restriction in [KomaMRIPlots/Project.toml](https://github.com/JuliaHealth/KomaMRI.jl/blob/master/KomaMRIPlots%2FProject.toml#L27) to PlutoPlotly 0.5, in case the users or collaborators create new notebooks:
```julia
PlutoPlotly = "0.4, 0.5"
```

pvillacorta:
Deleting this, since #562 already solves the problem.

PR #562, opened 2025-04-07
Restrict PlutoPlotly to version 0.5

Stockless:
This PR have the intention to fix the issue #550 , Having the following steps to solve the issue:

1. Restrict PlutoPlotly to version 0.5 at `KomaMRIPlots/Project.toml`
2. Release a new KomaMRI version (0.9.2) with the restriction from step 1
3. Update Reproductible-tutorials notebooks to use the new KomaMRI version
4. Release a new KomaMRI version (0.9.3) with the update from step 3

After all those steps, this will solve the issue #559 that is being solved by @pvillacorta in the PR #560 .
This solution will fix not only the currently existing notebooks, but also the notebooks that could be created by the Koma users and collaborators

[juliahealth.org/KomaMRI.jl/previews/PR562](https://juliahealth.org/KomaMRI.jl/previews/PR562/)


pvillacorta:
Hi! This PR and  #560 will not actually solve #559. These are temporal solutions until we find why PlutoPlotly v0.6 is not working properly. 
I would also delete this PR, and directly add the commit to #560 (since it is a related commit)

Issue #207, opened 2023-10-26
Mathematica source code missing in KomaMRI repo


beorostica:
We are performing simulation tests with results obtained from Mathematica simulations. It is a good idea to have the Mathematica source code stored in the KomaMRI repo.

cncastillo:
This should be replaced by DiffEq.jl.

Stockless:
In `KomaMRICore/test/runtest.jl:270` says that there is a Mathematica-simulated Bloch equation result being used in that test. What should I do about it?

cncastillo:
Can you link the code?

cncastillo:
Ok, just looked at it. This hardcoded results:

```julia
    #Mathematica-simulated Bloch equation result
    res1 = [0.153592+0.46505im,
            0.208571+0.437734im,
            0.259184+0.40408im,
            0.304722+0.364744im,
            0.344571+0.320455im,
            0.378217+0.272008im]
    res2 = [-0.0153894+0.142582im,
            0.00257641+0.14196im,
            0.020146+0.13912im,
            0.037051+0.134149im,
            0.0530392+0.12717im,
            0.0678774+0.11833im]
```

Should be replaced by something similar to the tests in Motion, which uses OrdinaryDiffEqTsit5.jl. This would be easier if #478 was solved, but it should be generated using DiffEq. 

I found a slight problem @pvillacorta , can you check if that is the case elsewhere?
- [ ] GPUFunctions tests: Line 416 shouldn't be "commented" (`@test true`)

Stockless:
Taking the Motion test as reference, do I also have to create a testset for every simulation method or keep using only the Bloch method?

```julia
for sim_method in [KomaMRICore.Bloch(), KomaMRICore.BlochSimple(), KomaMRICore.BlochDict()]
@testset "$(typeof(sim_method))" begin
...
```

cncastillo:
No, just look at what is inside, to see how to use DiffEq to get a solution.

Stockless:
created a starting point for this test update in a new branch, [commit](https://github.com/JuliaHealth/KomaMRI.jl/commit/9af1d1cf84fc2a46314b524ec17038b975587b4a)

PR #569, opened 2025-05-19
Bloch_RF_accuracy test reference update

Stockless:
Replaced the results reference from the Bloch_RF_accuracy test that was hardcoded with the generation of this result using OrdinaryDiffEqTsit5.

This Pull Request is generated from the issue mentioned at #207


cncastillo (approved):
Looks good! I cannot see on my phone, but make sure the issue is linked. What is the nmrse?

We can merge this if that is ok.

Stockless:
> Looks good! I cannot see on my phone, but make sure the issue is linked. 

The issue is linked correctly.

It is important to consider that this pull request solves an issue presented in the conversation of #207 that you mentioned. If that was the remaining problem in the issue, #207 should be considered closed.

> What is the nmrse?

NRMSE(raw, mxy_diffeq) = 0.3244598557717778

cncastillo:
Something is off, it should be less than 0.1%>

cncastillo:
Let's make sure tests pass, and I will merge!

cncastillo (changes requested):
Can you change the duration to `dur(seq)`?

Issue #571, opened 2025-05-30
Including the ADC frequency offset in the phase compensation


Stockless:
ADC frequency offset is not being used in the handling of the ADC phase offset in the `get_adc_phase_compensation` function.

The proposed ideal solution steps are:

1. Apply oversampling to the ADC data by calculating additional intermediate
ADC samples (e.g. oversampling by a factor of 2, that should not be too
expensive computationally);
2. apply phase modulation in accordance with the frequency and phase settings
of the ADC object;
3. apply pass-band filtering similar to what is done by the hardware;
4. compensate for the imperfect filter response and discard oversampling.

As a first approach, the recommendation is to implement the step 2 alone, adding a factor `exp(-i * 2π * freq_offset * t)`, where `t` is the timing of the sample relative to the start of the ADC object.

Thanks to @m-a-x-i-m-z for taking the time to tell us about this error and the previously presented proposed solution.

PR #572, opened 2025-06-02
Use Δf to calculate ADC phase compensation

Stockless:
This PR solves the issue #571 by updating the method used for ADC phase compensation, incorporating the frecuency offset.


cncastillo:
I think this is not the best solution, I believe I landed on the correct solution here: #657.

Issue #555, opened 2025-03-31
Pulseq 1.5.0


aTrotier:
### Feature Request

Pulseq 1.5.0 has been released (in matlab). KomaMRI does not handle this version

Maybe we can just add a warning for version higher to a specific version

cncastillo:
Have you encountered any problems using Pulseq 1.5.0? I think most of it should work.

rocthered:
> Have you encountered any problems using Pulseq 1.5.0? I think most of it should work.

I checked, the format pulseq store each event (gradient or RF) is different. That makes loading failed or wrong, depending on the sequence.

Issue #613, opened 2025-09-17
Some recon problem, probably due to definition of refocus pulse.


rocthered:
I have tried my variable-flip-angle (VFA) TSE sequence with KomaMRI. The default recon with MRIreco looked weird, especially when using large crushing gradients.

After some tests, I believe (although not 100% sure), the problem may be related to your definition of refocusing pulses. I found in the function get_RF_types:

	RF_ex = (α .<= 90.01) .* RF_mask
	RF_rf = (α .>  90.01) .* RF_mask

Seemed that when the flip angle is smaller than 90 deg, it will not be recognized as a refocusing pulse. However, in my VFA design, some refocusing flip angles are as low as ~20 deg. Maybe the code did not recognize them as refocusing pulses and misinterpreted the k-space trajectory. I can see that the k-space trajectory is wrong by plotting it.

In pulseq, the type of pulse can be manually set with "use". So when I plotted the trajectory from calculateKspacePP, it looked correct. 

If that is the case, can we get the type of pulses directly from pulseq, for pulseq sequences?

cncastillo:
Hi! Yes, now we're guessing as they did in older versions of pulseq. This new PR led by @Stockless will incorporate that information, and we can fix the kspace calculation!

If you have a pulseq that we could use to be sure we're reading it properly that would be very helpful. Maybe some pictures of how the kspace should look and how it looks now could help as well.

rocthered:
OK here's a detailed explanation of my problem.

I am using a variable flip angle sequence:

<img width="880" height="315" alt="Image" src="https://github.com/user-attachments/assets/5b2acb34-b4b7-43db-8e91-1224961414ed" />

You can see there are some very low-flip-angle refocusing pulses (especially around 0.05 ~ 0.15s).

However, in pulseq, I defined them as refocusing pulses:

[rfref] = mr.makeBlockPulse(flipref(echo_idx), 'Duration', 1e-3, 'system', system,'PhaseOffset',rfref_phase,'delay',rfref.delay+(rfref.shape_dur-1e-3)/2,'use','refocusing');

Therefore, pulseq gave me the correct kspace traj:

[ktraj_adc, t_adc, ktraj, t_ktraj, t_excitation, t_refocusing] = seq.calculateKspacePP();
figure; plot(ktraj(1,:),ktraj(2,:),'b',...
                 ktraj_adc(1,:),ktraj_adc(2,:),'r.'); 

[...]

Issue #575, opened 2025-06-05
Make extensions extensible


Stockless:
In the current implementation, extensions are being handled by using a long `if`/`elseif`chain, which is not very extensible, since new extensions requires modifying this block.
As @cncastillo mentioned in PR #558 , it would be good to replace this handling with multiple dispatch, allowing the users to define their own extensions without having to modify this code.

cncastillo:
Please give more details.

Stockless:
Sure! In PR #558 you made a comment on this lines of code (file KomaMRIFiles/src/Sequence/Pulseq.jl, lines 802-818)
```
    while true
        if extensionType[type]["data"] == "LABELSET"
            push!(EXT[1],LabelSet(labelsetLibrary[ref]["data"]...))
        elseif extensionType[type]["data"] == "LABELINC"
            push!(EXT[1],LabelInc(labelincLibrary[ref]["data"]...))
        elseif extensionType[type]["data"] == "TRIGGERS"
            push!(EXT[1],Trigger(triggerLibrary[ref]["data"]...))
        else
            @warn "Extension type not implemented"
        end

        if next_id == 0
            break
        else
            type, ref, next_id = extensionLibrary[next_id]["data"]
        end
    end
```
the comment said:
>It would be nice to do this with multiple dispatch, as extensions are meant to be extended. So if the user creates a new extension called "MyEXT" they can just define maybe read_extenstion(..., Val("MyEXT")), struct MyExT, or other functions like that, so they don't need to modify the code.

Issue #362, opened 2024-04-03
Create types for MRIEvents to enable multidispatch


beorostica:
Please view the comments:
* https://github.com/JuliaHealth/KomaMRI.jl/pull/321#issuecomment-2029950389
* https://github.com/JuliaHealth/KomaMRI.jl/pull/321#discussion_r1540162977

beorostica:
This is the most relevant comment:

Just to help, to dispatch based on the `Grad` type we need to redefine it like so:

```julia
using Parameters
abstract type MRISequenceEvent end
NumOrVect = Union{T, Vector{T}} where {T<:Number}

@with_kw_noshow mutable struct Grad{GradAmp<:NumOrVect, TimeDur<:NumOrVect} <: MRISequenceEvent
    # Grad properties
    A    ::GradAmp = zero(eltype(GradAmp))
    T    ::TimeDur = zero(eltype(TimeDur))
    rise ::eltype(TimeDur) = zero(eltype(T))
    fall ::eltype(TimeDur) = rise
    delay::eltype(TimeDur) = zero(eltype(T))
    first::eltype(GradAmp) = zero(eltype(A))
    last ::eltype(GradAmp) = first
    # Checking validity of Grad
    @assert length(A) >= length(T)
    @assert all(T .>= zero(eltype(T)))
    @assert rise   >= zero(eltype(T))
    @assert fall   >= zero(eltype(T))
    @assert delay  >= zero(eltype(T))
end

# Gradient type aliases for dispatching
TrapezoidalGrad      = Grad{G, T} where {G, T}
UniformlySampledGrad = Grad{Vector{G}, T} where {G, T}
TimeShapedGrad       = Grad{Vector{G}, Vector{T}} where {G, T}
ValidGrad = Union{TrapezoidalGrad{G, T}, UniformlySampledGrad{G, T}, TimeShapedGrad{G, T}} where {G, T}

# MRISequenceEvent functions
# Event duration
dur(g::Grad) = g.delay + g.rise + sum(g.T) + g.fall
dur(G::Vector{Grad{G, T}}) where {G, T} = max(dur.(G)...)
# Event amplitude
ampl(g::TrapezoidalGrad{G,T})      where {G, T} = vcat(zero(G), g.first, g.A, g.A, g.last)
ampl(g::TimeShapedGrad{G,T})       where {G, T} = vcat(zero(G), g.first, g.A, g.last)
[...]

PR #614, opened 2025-09-17
Read pulseq 1.5.0

Stockless:
This PR addresses issue #555 by adding support for reading Pulseq files in version 1.5.0.

### DIFF
--- a/KomaMRIFiles/src/Sequence/Pulseq.jl
+++ b/KomaMRIFiles/src/Sequence/Pulseq.jl
@@ -366,14 +366,18 @@ function read_seq(filename)
                 end
                 blockEvents, blockDurations, delayInd_tmp = read_blocks(io, def["BlockDurationRaster"], version_combined)
             elseif  section == "[RF]"
-                if version_combined >= 1004000
+                if version_combined >= 1005000
+                    rfLibrary = read_events(io, [1/γ 1 1 1 1 1e-6 1 1 1 1 1]) # this is 1.5.x format
+                elseif version_combined >= 1004000
                     rfLibrary = read_events(io, [1/γ 1 1 1 1e-6 1 1]) # this is 1.4.x format
                 else
                     rfLibrary = read_events(io, [1/γ 1 1 1e-6 1 1]) # this is 1.3.x and below
                     # we will have to scan through the library later after all the shapes have been loaded
                 end
             elseif  section == "[GRADIENTS]"
-                if version_combined >= 1004000
+                if version_combined >= 1005000
+                    gradLibrary = read_events(io, [1/γ 1 1 1 1 1e-6]; type='g', eventLibrary=gradLibrary) # this is 1.5.x format
+                elseif version_combined >= 1004000
                     gradLibrary = read_events(io, [1/γ 1 1 1e-6]; type='g', eventLibrary=gradLibrary) # this is 1.4.x format
                 else
                     gradLibrary = read_events(io, [1/γ 1 1e-6];   type='g', eventLibrary=gradLibrary) # this is 1.3.x and below
@@ -381,7 +385,11 @@ function read_seq(filename)
             elseif  section == "[TRAP]"
                 gradLibrary = read_events(io, [1/γ 1e-6 1e-6 1e-6 1e-6]; type='t', eventLibrary=gradLibrary);
             elseif  section == "[ADC]"
-                adcLibrary = read_events(io, [1 1e-9 1e-6 1 1])
+                if version_combined >= 1005000
+                    adcLibrary = read_events(io, [1 1e-9 1e-6 1 1 1 1 1]) # this is 1.5.x format
+                else
+                    adcLibrary = read_events(io, [1 1e-9 1e-6 1 1]) # this is 1.4.x and below
+                end
             elseif  section == "[DELAYS]"
                 if version_combined >= 1004000
                     @error "Pulseq file revision 1.4.0 and above MUST NOT contain [DELAYS] section"
@@ -455,10 +463,13 @@ function read_seq(filename)
     #This should only work for Pulseq files >=1.4.0
     seq = Sequence()
     for i = 1:length(blockEvents)
-        seq += get_block(obj, i)
+        seq += get_block(obj, i, version_combined)
     end
-    # Add first and last points for gradients #320
-    fix_first_last_grads!(seq)
+    # Add first and last points for gradients #320 for version <= 1.4.2
+    if version_combined < 1005000
+        @warn "Changing first and last points for gradients"
+        fix_first_last_grads!(seq)
+    end
     # Final details
     # Hack for including extension and triggers, this will be done properly for #308 and #323
     seq.DEF["additional_text"] = read_Extension(extensionLibrary, triggerLibrary) #Temporary hack
@@ -486,11 +497,12 @@ Reads the gradient. It is used internally by [`get_block`](@ref).
 - `shapeLibrary`: (`::Dict{K, V}`) the "shapeLibrary" dictionary
 - `Δt_gr`: (`::Float64`, `[s]`) gradient raster time
 - `i`: (`::Int64`) index of the axis in the block event
+- `version_combined`: (`::Int64`) version of the Pulseq file
 
 # Returns
 - `grad`: (::Grad) Gradient struct
 """
-function read_Grad(gradLibrary, shapeLibrary, Δt_gr, i)
+function read_Grad(gradLibrary, shapeLibrary, Δt_gr, i, version_combined)
     G = Grad(0.0,0.0)
     if isempty(gradLibrary) || i==0
         return G
@@ -501,12 +513,23 @@ function read_Grad(gradLibrary, shapeLibrary, Δt_gr, i)
         g_A, g_rise, g_T, g_fall, g_delay = gradLibrary[i]["data"]
         G = Grad(g_A,g_T,g_rise,g_fall,g_delay)
     elseif gradLibrary[i]["type"] == 'g' #Arbitrary gradient waveform
-        #(1)amplitude (2)amp_shape_id (3)time_shape_id (4)delay
-        g = gradLibrary[i]["data"]
-        amplitude =     g[1]
-        amp_shape_id =  g[2] |> x->floor(Int64,x)
-        time_shape_id = g[3] |> x->floor(Int64,x)
-        delay =         g[4]
+        if version_combined >= 1005000
+            #(1)amplitude (2)first_grads (3)last_grads (4)amp_shape_id (5)time_shape_id (6)delay
+            g = gradLibrary[i]["data"]
+            amplitude =     g[1]
+            first_grads =   g[2]
+            last_grads =    g[3]
+            amp_shape_id =  g[4] |> x->floor(Int64,x)
+            time_shape_id = g[5] |> x->floor(Int64,x)
+            delay =         g[6]
+        else
+            #(1)amplitude (2)amp_shape_id (3)time_shape_id (4)delay
+            g = gradLibrary[i]["data"]
+            amplitude =     g[1]
+            amp_shape_id =  g[2] |> x->floor(Int64,x)
+            time_shape_id = g[3] |> x->floor(Int64,x)
+            delay =         g[4]
+        end
         #Amplitude
         gA = amplitude * decompress_shape(shapeLibrary[amp_shape_id]...)
         Nrf = length(gA) - 1
@@ -533,26 +556,42 @@ Reads the RF. It is used internally by [`get_block`](@ref).
 - `shapeLibrary`: (`::Dict{K, V}`) the "shapeLibrary" dictionary
 - `Δt_rf`: (`::Float64`, `[s]`) RF raster time
 - `i`: (`::Int64`) index of the RF in the block event
+- `version_combined`: (`::Int64`) version of the Pulseq file
 
 # Returns
 - `rf`: (`1x1 ::Matrix{RF}`) RF struct
 """
-function read_RF(rfLibrary, shapeLibrary, Δt_rf, i)
+function read_RF(rfLibrary, shapeLibrary, Δt_rf, i, version_combined)
 
     if isempty(rfLibrary) || i==0
         return reshape([RF(0.0,0.0)], 1, 1)
     end
 
     #Unpacking
-    #(1)amplitude (2)mag_id (3)phase_id (4)time_shape_id (5)delay (6)freq (7)phase
     r = rfLibrary[i]["data"]
-    amplitude =     r[1]
-    mag_id =        r[2] |> x->floor(Int64,x)
-    phase_id =      r[3] |> x->floor(Int64,x)
-    time_shape_id = r[4] |> x->floor(Int64,x)
-    delay =         r[5] + (time_shape_id==0)*Δt_rf/2
-    freq =          r[6]
-    phase =         r[7]
+    if version_combined >= 1005000
+        #(1)amplitude (2)mag_id (3)phase_id (4)time_shape_id (5)center (6)delay (7)freq_ppm (8)phase_ppm (9)freq (10)phase (11)use
+        amplitude =     r[1]
+        mag_id =        r[2] |> x->floor(Int64,x)
+        phase_id =      r[3] |> x->floor(Int64,x)
+        time_shape_id = r[4] |> x->floor(Int64,x)
+        center =        r[5]
+        delay =         r[6] + (time_shape_id==0)*Δt_rf/2
+        freq_ppm =      r[7]
+        phase_ppm =     r[8]
+        freq =          r[9]
+        phase =         r[10]
+        use =           r[11]
+    else
+        #(1)amplitude (2)mag_id (3)phase_id (4)time_shape_id (5)delay (6)freq (7)phase
+        amplitude =     r[1]
+        mag_id =        r[2] |> x->floor(Int64,x)
+        phase_id =      r[3] |> x->floor(Int64,x)
+        time_shape_id = r[4] |> x->floor(Int64,x)
+        delay =         r[5] + (time_shape_id==0)*Δt_rf/2
+        freq =          r[6]
+        phase =         r[7]
+    end
     #Amplitude and phase waveforms
     if amplitude != 0 && mag_id != 0
         rfA = decompress_shape(shapeLibrary[mag_id]...)
@@ -582,24 +621,38 @@ Reads the ADC. It is used internally by [`get_block`](@ref).
 # Arguments
 - `adcLibrary`: (`::Dict{String, Any}`) the "adcLibrary" dictionary
 - `i`: (`::Int64`) index of the adc in the block event
+- `version_combined`: (`::Int64`) version of the Pulseq file
 
 # Returns
 - `adc`: (`1x1 ::Vector{ADC}`) ADC struct
 """
-function read_ADC(adcLibrary, i)
+function read_ADC(adcLibrary, i, version_combined)
 
     if isempty(adcLibrary) || i==0
         return [ADC(0, 0)]
     end
 
     #Unpacking
-    #(1)num (2)dwell (3)delay (4)freq (5)phase
-    a = adcLibrary[i]["data"]
-    num =   a[1] |> x->floor(Int64,x)
-    dwell = a[2]
-    delay = a[3] + dwell/2
-    freq =  a[4]
-    phase = a[5]
+    if version_combined >= 1005000
+        #(1)num (2)dwell (3)delay (4)freq_ppm (5)phase_ppm (6)freq (7)phase (8)phase_shape_id
+        a = adcLibrary[i]["data"]
+        num =   a[1] |> x->floor(Int64,x)
+        dwell = a[2]
+        delay = a[3] + dwell/2
+        freq_ppm = a[4]
+        phase_ppm = a[5]
+        freq =  a[6]
+        phase = a[7]
+        phase_shape_id = a[8] |> x->floor(Int64,x)
+    else
+        #(1)num (2)dwell (3)delay (4)freq (5)phase
+        a = adcLibrary[i]["data"]
+        num =   a[1] |> x->floor(Int64,x)
+        dwell = a[2]
+        delay = a[3] + dwell/2
+        freq =  a[4]
+        phase = a[5]
+    end
     #Definition
     T = (num-1) * dwell
     A = [ADC(num,T,delay,freq,phase)]
@@ -681,27 +734,28 @@ Block sequence definition. Used internally by [`read_seq`](@ref).
 # Arguments
 - `obj`: (`::Dict{String, Any}`) main dictionary
 - `i`: (`::Int64`) index of a block event
+- `version_combined`: (`::Int64`) version of the Pulseq file
 
 # Returns
 - `s`: (`::Sequence`) block Sequence struct
 """
-function get_block(obj, i)
+function get_block(obj, i, version_combined)
     #Unpacking
     idelay, irf, ix, iy, iz, iadc, iext = obj["blockEvents"][i]
 
     #Gradient definition
     Δt_gr = obj["definitions"]["GradientRasterTime"]
-    Gx = read_Grad(obj["gradLibrary"], obj["shapeLibrary"], Δt_gr, ix)
-    Gy = read_Grad(obj["gradLibrary"], obj["shapeLibrary"], Δt_gr, iy)
-    Gz = read_Grad(obj["gradLibrary"], obj["shapeLibrary"], Δt_gr, iz)
+    Gx = read_Grad(obj["gradLibrary"], obj["shapeLibrary"], Δt_gr, ix, version_combined)
+    Gy = read_Grad(obj["gradLibrary"], obj["shapeLibrary"], Δt_gr, iy, version_combined)
+    Gz = read_Grad(obj["gradLibrary"], obj["shapeLibrary"], Δt_gr, iz, version_combined)
     G = reshape([Gx;Gy;Gz],3,1) #[Gx;Gy;Gz;;]
 
     #RF definition
     Δt_rf = obj["definitions"]["RadiofrequencyRasterTime"]
-    R = read_RF(obj["rfLibrary"], obj["shapeLibrary"], Δt_rf, irf)
+    R = read_RF(obj["rfLibrary"], obj["shapeLibrary"], Δt_rf, irf, version_combined)
 
     #ADC definition
-    A = read_ADC(obj["adcLibrary"], iadc)
+    A = read_ADC(obj["adcLibrary"], iadc, version_combined)
 
     #DUR
     D = Float64[max(obj["blockDurations"][i], dur(Gx), dur(Gy), dur(Gz), dur(R[1]), dur(A[1]))]


KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -366,22 +366,30 @@ function read_seq(filename)
                 end
                 blockEvents, blockDurations, delayInd_tmp = read_blocks(io, def["BlockDurationRaster"], version_combined)
             elseif  section == "[RF]"
-                if version_combined >= 1004000
```
cncastillo: Don't remove the 1.4 reader

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -366,22 +366,30 @@ function read_seq(filename)
...
+                    rfLibrary = read_events(io, [1/γ 1 1 1 1 1e-6 1 1 1 1 1]) # this is 1.5.x format
+                elseif version_combined >= 1004000
                     rfLibrary = read_events(io, [1/γ 1 1 1 1e-6 1 1]) # this is 1.4.x format
                 else
                     rfLibrary = read_events(io, [1/γ 1 1 1e-6 1 1]) # this is 1.3.x and below
                     # we will have to scan through the library later after all the shapes have been loaded
                 end
             elseif  section == "[GRADIENTS]"
-                if version_combined >= 1004000
+                if version_combined >= 1005000
```
cncastillo: Same here.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -366,22 +366,30 @@ function read_seq(filename)
...
+                elseif version_combined >= 1004000
                     gradLibrary = read_events(io, [1/γ 1 1 1e-6]; type='g', eventLibrary=gradLibrary) # this is 1.4.x format
                 else
                     gradLibrary = read_events(io, [1/γ 1 1e-6];   type='g', eventLibrary=gradLibrary) # this is 1.3.x and below
                 end
             elseif  section == "[TRAP]"
                 gradLibrary = read_events(io, [1/γ 1e-6 1e-6 1e-6 1e-6]; type='t', eventLibrary=gradLibrary);
             elseif  section == "[ADC]"
-                adcLibrary = read_events(io, [1 1e-9 1e-6 1 1])
+                if version_combined >= 1005000
```
cncastillo: Same here

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -455,10 +463,13 @@ function read_seq(filename)
     #This should only work for Pulseq files >=1.4.0
     seq = Sequence()
     for i = 1:length(blockEvents)
-        seq += get_block(obj, i)
+        seq += get_block(obj, i, version_combined)
+    end
+    # Add first and last points for gradients #320 for version <= 1.4.2
+    if version_combined < 1005000
+        @warn "Changing first and last points for gradients"
```
cncastillo: Why did you put a warning here

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -533,26 +556,42 @@ Reads the RF. It is used internally by [`get_block`](@ref).
...
+        mag_id =        r[2] |> x->floor(Int64,x)
+        phase_id =      r[3] |> x->floor(Int64,x)
+        time_shape_id = r[4] |> x->floor(Int64,x)
+        center =        r[5]
+        delay =         r[6] + (time_shape_id==0)*Δt_rf/2
+        freq_ppm =      r[7]
+        phase_ppm =     r[8]
+        freq =          r[9]
+        phase =         r[10]
+        use =           r[11]
```
cncastillo: You are not using `use`, there's an issue that could be solved by incorporating this information into the RF #613.

cncastillo (changes requested):
Please use three correct logic so pulse 1.4 still works. No need to pass the version into the get_X functions, you can just check the input length of the library.

KomaMRIFiles/src/Sequence/Pulseq.jl
Stockless: I'm not removing it, that version is 2 lines bellow (I added the version 1.5.x as the first case and 1.4.x second, to keep them sorted)

KomaMRIFiles/src/Sequence/Pulseq.jl
Stockless: Removed at [54bb7e1](https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/54bb7e154211e2e4b3cea3a278a21d19b07807e7)

KomaMRIFiles/src/Sequence/Pulseq.jl
Stockless: Solved at [54bb7e1](https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/54bb7e154211e2e4b3cea3a278a21d19b07807e7)

KomaMRIFiles/src/Sequence/Pulseq.jl
Stockless: Solved at [54bb7e1](https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/54bb7e154211e2e4b3cea3a278a21d19b07807e7)

cncastillo:
Modify to consider new RF uses, to calculate k-space correctly (not guess RF use by > 90.0):

- get_RF_types() → 1st
- get_RF_center() → 2nd
- get_Mk() → 3rd

If use is Undefined(), keep using the current method; otherwise, use the provided value.

If center is not defined (i.e., == 0.0), keep the current method; otherwise, use the provided value.

Test with reader with (take inspiration from https://juliahealth.org/KomaMRI.jl/dev/explanation/5-seq-events/):
- arbitrary gradients
- time shaped gradients
- trapezoidal gradients
- arbitrary rfs
- time shaped rfs
- block rfs 

@Stockless can you give use the MATLAB files you were using for the tests?

Update get_events to add `compat_helper` copy: https://github.com/pulseq/pulseq/blob/master/matlab/%2Bmr/%40Sequence/read.m#L520

pvillacorta:
> @pvillacorta , @cncastillo mentioned you are interested in working this PR out. I want to get it done too. Let's meet to sort out the efforts?

Yes! Let's join forces :)

cncastillo:
@pvillacorta @gsahonero do you guys want to plan a meeting to push this?

pvillacorta:
Pulseq files from `KomaMRIFiles/test/test_files/pulseq_read_comparison` must ve reviewed, since I have not been able to retrieve their signature. For the moment, I have just added a warning when the retrieved signature and the expected one do not coincide (obviously, the ideal case would be to throw an error, but tests would not pass for `pulseq_read_comparison`).

KomaMRIBase/src/datatypes/sequence/RF.jl
```diff
@@ -1,3 +1,31 @@
...
+)
+
+get_RF_use_from_char(c::Char) = RF_USE_CHARS[c]
+get_RF_use_from_char(u::RFUse) = u
+get_char_from_RF_use(use::RFUse) = begin
+    for (k, v) in RF_USE_CHARS
+        if v == use return k end
+    end
+    return 'u'
+end
```
cncastillo: Why did you change this? `Val{use}()` was used so the user can define new `use`s easily. Or Pulseq breaks if another character is used? We are on purpose expanding the name from `'u'` to `Undefined()` so people actually know what they mean without checking the docs.

KomaMRIBase/test/runtests.jl
```diff
@@ -150,11 +150,11 @@ using TestItems, TestItemRunner
 
         #Sanity checks of constructors (A [T], T [s], Δf[Hz], delay [s])
         A, T = rand(2)
-        r1, r2 = RF(A,T), RF(A,T,0.0,0.0)
+        r1, r2 = RF(A,T), RF(A,T,0.0,0.0,0.0,'u')
```
cncastillo: using `'u'` instead of `Undefined()::RFUse` is probably not correct, that field must be of type `RFUse`

KomaMRIFiles/test/runtests.jl
```diff
@@ -11,22 +11,18 @@ using TestItems, TestItemRunner, KomaMRIBase
         seq = @suppress read_seq(pth*"/test_files/epi.seq") #Pulseq v1.4.0, RF arbitrary
         @test seq.DEF["FileName"] == "epi.seq"
         @test seq.DEF["PulseqVersion"] == v"1.4.0"
-        @test seq.DEF["signature"] == "67ebeffe6afdf0c393834101c14f3990"
```
cncastillo, line 14: why did you remove these?

KomaMRIBase/src/datatypes/sequence/RF.jl
```diff
@@ -28,19 +56,13 @@ mutable struct RF
...
-    function RF(A, T)
-        return any(T .< 0) ? error("RF timings must be non-negative.") : new(A, T, 0.0, 0.0)
-    end
+    center
+    use::RFUse
+    RF(A, T, Δf, delay, center, use) = any(T .< 0) || delay < 0 ? error("RF timings must be non-negative.") : new(A, T, Δf, delay, center, get_RF_use_from_char(use))
+    RF(A, T, Δf, delay, center) = RF(A, T, Δf, delay, center, Undefined())
+    RF(A, T, Δf, delay) = RF(A, T, Δf, delay, 0.0)
+    RF(A, T, Δf) = RF(A, T, Δf, 0.0)
+    RF(A, T) = RF(A, T, 0.0)
```
cncastillo: Remove the use of `get_RF_use_from_char` in the constructor.

Don't use recursive definitions. 
```suggestion
    center
    use::RFUse
    RF(A, T, Δf, delay, center, use) = any(T .< 0) || delay < 0 ? error("RF timings must be non-negative.") : new(A, T, Δf, delay, center, use)
    RF(A, T, Δf, delay, center) = RF(A, T, Δf, delay, center, Undefined())
    RF(A, T, Δf, delay) = RF(A, T, Δf, delay, 0.0, Undefined())
    RF(A, T, Δf) = RF(A, T, Δf, 0.0, 0.0, Undefined())
    RF(A, T) = RF(A, T, 0.0, 0.0, 0.0, Undefined())
```

KomaMRIBase/src/datatypes/sequence/Grad.jl
```diff
@@ -83,34 +83,11 @@ mutable struct Grad
...
-            error("Gradient timings must be positive.")
-        else
-            new(A, T, 0.0, 0.0, 0.0, 0.0, 0.0)
-        end
-    end
+    Grad(A, T, rise, fall, delay, first, last) = all(T .< 0) || rise < 0 || fall < 0 || delay < 0 || first < 0 || last < 0 ? error("Gradient timings must be positive.") : new(A, T, rise, fall, delay, first, last)
+    Grad(A, T, rise, fall, delay) = Grad(A, T, rise, fall, delay, 0.0, 0.0)
+    Grad(A, T, rise, delay) = Grad(A, T, rise, rise, delay)
+    Grad(A, T, rise) = Grad(A, T, rise, rise, 0.0)
+    Grad(A, T) = Grad(A, T, 0.0)
```
cncastillo: Don't use recursive definitions.

```suggestion
    Grad(A, T, rise, fall, delay, first, last) = all(T .< 0) || rise < 0 || fall < 0 || delay < 0 || first < 0 || last < 0 ? error("Gradient timings must be positive.") : new(A, T, rise, fall, delay, first, last)
    Grad(A, T, rise, fall, delay) = Grad(A, T, rise, fall, delay, 0.0, 0.0)
    Grad(A, T, rise, delay) = Grad(A, T, rise, rise, delay, 0.0, 0.0)
    Grad(A, T, rise) = Grad(A, T, rise, rise, 0.0, 0.0, 0.0)
    Grad(A, T) = Grad(A, T, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
```

KomaMRIFiles/test/runtests.jl
pvillacorta: Signature verification is now done inside the `read_seq` function. Until now, no verification was done: signature string was just stored in seq.DEF, and then compared with "itself" (obviously, this would always return true). Now, signature is computed from the .seq file payload, and compared with the one contained in the file. Besides, I do not consider necessary to store the signature within the `Sequence` struct, since it only needs to be computed from the `read_seq` and `write_seq` functions. I think it would be confusing to store the signature, actually.

aTrotier: A quick remark, I don't have the time to dig further into it so maybe it is already the case.

Please just use a warning if the signature stored in the file is not the same as the one computed. Sometimes it is useful to just change some values by hand in a .seq file.

pvillacorta: Hi! Yes, it is the case now. A warning is thrown when signatures do not match. In fact, sequences from `KomaMRIFiles/test/test_files/pulseq_read_comparison` might have been manually modified, since extracted and computed signatures do not match. The warning (and not an error) is thrown to make the tests pass.

KomaMRIBase/src/datatypes/sequence/RF.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/c901565a9c81b6d5513be154000bd32bfb90ae97

KomaMRIBase/test/runtests.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/c901565a9c81b6d5513be154000bd32bfb90ae97

KomaMRIBase/src/datatypes/sequence/RF.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/c901565a9c81b6d5513be154000bd32bfb90ae97

KomaMRIBase/src/datatypes/sequence/Grad.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/c901565a9c81b6d5513be154000bd32bfb90ae97

KomaMRIFiles/test/runtests.jl
cncastillo: This test is not to check if the signature matches, it is just to check if we read it correctly. Please put them back. I agree with the warning.

pvillacorta: Solved in b308bebfff131d2b581acf8c2895b5b52755a202

pvillacorta:
I have modified and created new MATLAB scripts into a local version of [this folder](https://github.com/pulseq/pulseq/tree/master/matlab/demoSeq) in order to make similar (v1.5.1) sequences to those existing in v1.4.
Testing aspects such as labels and extensions is pending.

pvillacorta:
The commit above tries to address #575. The main change is that it is no longer necessary to create a library for each extension type inside the reader. Now, extensions are managed through three main dictionaries within the`read_seq` function: `extensionInstanceLibrary`, `extensionTypeLibrary`, and `extensionSpecLibrary`. 
If we wanted to define a new extension, we would just need to create its corresponding file within the `KomaMRIBase/src/datatypes/sequence/extensions` directory (obviously, if we wanted that extension to take effect, for instance, in the simulation, more changes would need to be made).
The following [v1.5.1 extensions](https://pulseq.github.io/specification.pdf#subsubsection.2.8.4) are not implemented yet:
- [ ] Soft Delay
- [ ] Rotation
- [ ] RF Shimming

Maybe leave them for another PR? 
A hint about how to do this for the Soft Delay case (not implemented yet):
`KomaMRIBase/src/datatypes/sequence/extensions/SoftDelay.jl`:
```julia
mutable struct SoftDelay <: Extension
    num::Int
    offset::Int
    factor::Int
    hint::String
end
```
Let me know what you think, I hope we can merge this soon in order to also address https://github.com/JuliaHealth/KomaMRI.jl/issues/152 during the following days.

KomaMRIBase/src/datatypes/sequence/extensions/LabelInc.jl
```diff
@@ -0,0 +1,6 @@
+mutable struct LabelInc <: Extension
+    labelvalue::Int
+    labelstring::String
+end
+
+get_scale(::Type{LabelInc}) = [1 ""]
```
cncastillo: For performance, I would try to make the `get_scale` function to return a `Tuple` instead of an `Array{Any}`.

KomaMRIBase/src/datatypes/sequence/EXT.jl
```diff
@@ -1,60 +1,39 @@
...
-  d1::Float64 # Delay prior to the trigger event (us)
-  d2::Float64 # Duration of trigger event (us)
+# Supported extensions. To add a new extension, create a new file in the extensions folder and add it to the list below.
+include("extensions/LabelInc.jl")
+include("extensions/LabelSet.jl")
+include("extensions/Trigger.jl")
+
+get_EXT_type_from_symbol(::Val{:LABELINC}) = LabelInc
+get_EXT_type_from_symbol(::Val{:LABELSET}) = LabelSet
+get_EXT_type_from_symbol(::Val{:TRIGGERS}) = Trigger
```
cncastillo: I would define this in the extensions files, so if I want to create a new one I can know which functions to generate.

KomaMRIBase/src/datatypes/sequence/EXT.jl
```diff
@@ -1,60 +1,39 @@
...
+        Int16 => "%i",
+        Int8 => "%i",
+        Float64 => "%f",
+        Float32 => "%f",
+        String => "%s",
+        Char => "%c"
+    )
+    format_parts = [get(type_to_format, ft, "%s") for ft in fieldtypes(T)]
+    return join(format_parts, " ")
 end
```
cncastillo: I this might be overly complicating stuff, not sure why is needed. A format string per extension might be good enough, `get_scanf_format(::Trigger) = "%i %i %f %f"`. Is this only used by extensions?

cncastillo (changes requested):
I jut wanted to say, that the introduction of `trapLibrary` and other dictionaries seem to deviate too much from the MALTAB's Pulseq reader. I would like to follow the same logic, even if inefficient so it is easier to update when new versions are released.

Also, the whole handling of fmt strings for `scanf` (like the function `get_scanf_format`) seem overly complicated. The comment I left gives a solution, but there might be an easier one like:

```julia
struct MyStructure
	x::Float64
    y::Int
end
t = MyStructure(parse.(fieldtypes(MyStructure), split("1.0 1"))...)
```

The extensive use of `Dict`'s to map inputs to types seems weird to me, and all the generated arrays of type `Any` are not very performant.

Finally, I think there is too many functions like `f(x::Type{T})` instead of `f(x::T)` which seems too complicated.

I haven't reviewed the whole PR, but I don't think in the current state in can be merged quickly, there are too many "unnecessary" changes. For example, there are some breaking changes on how to read Pulseq <1.4.

My main point: Let's try to stick to whatever Pulseq is doing, let's only deviate for performance or extensibility.

KomaMRIBase/src/datatypes/sequence/extensions/LabelInc.jl
pvillacorta: Instead of Tuples, I just used Float64 arrays, since the `""` elements were actually unnecessary

KomaMRIBase/src/datatypes/sequence/EXT.jl
pvillacorta: Done

KomaMRIBase/src/datatypes/sequence/EXT.jl
pvillacorta: Done

pvillacorta:
> I jut wanted to say, that the introduction of `trapLibrary` and other dictionaries seem to deviate too much from the MALTAB's Pulseq reader. I would like to follow the same logic, even if inefficient so it is easier to update when new versions are released.

I agree that this deviates from the exact structure of the MATLAB reader. The main motivation for introducing `trapLibrary` was to simplify `read_events`, which was previously very hard-coded:

- RF events relied on character-level checks (e.g. last character determining RF type).
- Gradients were distinguished via tag characters (`t`vs `g`) scattered throughout the code.

By separating trapezoidal gradients from arbitrary gradients into a dedicated library, the dispatch logic in `read_events` becomes simpler and more explicit. This reduces the amount of branching and character-based parsing while still preserving the underlying Pulseq concepts.

At this point, I see this as a pragmatic compromise: it deviates slightly from MATLAB’s implementation, but significantly reduces code complexity and improves maintainability. My intention was not to redesign the Pulseq logic, but to make the Julia reader easier to reason about and extend while keeping the mapping to Pulseq concepts clear.

> Also, the whole handling of fmt strings for `scanf` (like the function `get_scanf_format`) seem overly complicated. The comment I left gives a solution, but there might be an easier one like:
> 
> ```julia
> struct MyStructure
> 	x::Float64
>     y::Int
> end
> t = MyStructure(parse.(fieldtypes(MyStructure), split("1.0 1"))...)
> ```

You’re absolutely right here, and I’ve already addressed this in the latest commit.

I removed the generic auxiliary functions (`get_scale`, `get_scanf_args`) and instead implemented the parsing logic directly in each extension and inside `read_events`. This removes unnecessary abstraction and makes the parsing much more straightforward and readable.

> The extensive use of `Dict`'s to map inputs to types seems weird to me, and all the generated arrays of type `Any` are not very performant.

This is a fair concern, especially regarding performance.

Here I intentionally tried to stay close to what was already present in the reader before this PR. Dictionaries were already used extensively, and keeping them allowed me to:

- Minimize the scope of changes.
- Stay conceptually close to Pulseq’s `EventsLibrary` (even though Pulseq uses a struct rather than a Dict).

I fully agree that performance could be improved by using more concrete types (or even arrays directly, as it was stated in #224). However, given time constraints and the goal of not overengineering this PR, I preferred to keep consistency with the existing codebase rather than introduce a larger refactor.

> Finally, I think there is too many functions like `f(x::Type{T})` instead of `f(x::T)` which seems too complicated.

The only places where I dispatch on `Type{T}` are `get_scanf_format` and `get_scale`. This was a deliberate choice to integrate the extension handling into`read_events` without introducing a separate reading pipeline. Dispatching on types allows `read_events` to treat extensions in the same way as other events (such as gradients or RFs).

Instantiating a dummy object before parsing its arguments felt less practical (and potentially more expensive) than dispatching on the type directly. While I agree this is not ideal in general, in this context it seemed like a reasonable trade-off to keep the overall logic simpler and more uniform.

> I haven't reviewed the whole PR, but I don't think in the current state in can be merged quickly, there are too many "unnecessary" changes. For example, there are some breaking changes on how to read Pulseq <1.4.

I understand this concern, but I’m not entirely sure which specific breaking changes for Pulseq < 1.4 you’re referring to. If you can point out a concrete example or file, I would try to adjust or revert that part.

> My main point: Let's try to stick to whatever Pulseq is doing, let's only deviate for performance or extensibility.

I fully agree with your main point. we should stick to Pulseq's logic and only deviate when there is a clear benefit in terms of performance or extensibility. My intention was exactly that, not to redesign the reader functions, but to make the Julia reader usable, maintainable and more extensible. Given the time constraints, I think this is a reasonable compromise, and the upcoming Pulseq writer PR will follow the same direction.
Thank you!

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -420,39 +420,34 @@ get_flip_angles(x::Sequence) = get_flip_angle.(x.RF)[:]
...
+			push!(rf_idx, argmin(abs.(trf.-t))...)
+			push!(rf_types, _get_RF_use(rf, rf.use))
 		end
 	end
-	rf_idx, rf_type
+	return rf_idx, rf_types
 end
 
+_get_RF_use(rf::RF, use::RFUse)  = use
+_get_RF_use(rf::RF, ::Undefined) = get_flip_angle(rf) <= 90.01 ? Excitation() : Refocusing()
```
cncastillo: While I kind of agree with the idea, It it kind of weird that this function needs two parameters. If we had `RF{RFUse}` we can just do

```julia
_get_RF_use(rf::RF) = rf.use
_get_RF_use(rf::RF{Undefined}) = get_flip_angle(rf) <= 90.01 ? Excitation() : Refocusing()
```

Also, maybe let's change the name from `_get_RF_use` to `_moment_calc_RFUse`. What does Pulseq do for k-space calculation?

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -480,19 +478,15 @@ function get_Mk(seq::Sequence, k; Δt=1, skip_rf=zeros(Bool, sum(is_RF_on.(seq))
...
 				if !skip_rf[rf-1]
-					if rf_type[rf-1] == 0 # Excitation
-						mk[p,i] .-= 0
-					elseif rf_type[rf-1] == 1 # Refocuse
-						mk[p,i] .-= mkf
-					end
+					mk[p,i] .+= mkf * get_sign(rf_types[rf-1])
 				else
 					mk[p,i] .+= mkf
 				end
```
cncastillo: The `skip_rf` is basically doing `rf.use = Other()` (or anything other than `Excitation`, `Refocusing`, and `Undefined`). I would remove it (marking KomaMRIFiles as 0.10, breaking change).

KomaMRIBase/src/timing/TimeStepCalculation.jl
```diff
@@ -102,7 +102,7 @@ function get_variable_times(seq; Δt=1e-3, Δt_rf=1e-5, motion=NoMotion())
 			delay, T = y.delay, y.T
 			t1 = t0 + delay
 			t2 = t1 + sum(T)
-			rf0 = t0 + get_RF_center(y) #get_RF_center includes delays
+			rf0 = t0 + y.center # y.center includes delays
```
cncastillo: Is this true? I thought the Pulseq's center was defined from the start of the RF pulse, not the start of the block. My calculation for the center includes the delay.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -52,25 +48,28 @@ function read_definitions(io)
...
-            signature = length(parsed_array) == 1 ? parsed_array[1] : parsed_array
+        value = join(line_split[2:end], " ")
+        if key == "Type"
+            signature_type = strip(value)
+        elseif key == "Hash"
+            signature_hash = lowercase(replace(strip(value), " " => ""))
         end
     end
-    return signature
+    return isnothing(signature_type) && isnothing(signature_hash) ? nothing : (type = signature_type, hash = signature_hash)
```
cncastillo: Put `isnothing(signature_type) && isnothing(signature_hash)` into a variable `has_hash` (?)

Why not return an empty string? what does Pulseq do?

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -114,69 +113,24 @@ end
...
+        r, data... = scanf(line, fmt, args...)
+        r == eventLength || break #Break if not all values read
         id = floor(Int, data[1])
-        eventLibrary[id] = Dict("data"=>data[2:4])
-        r == 4 || break #Break on white space
+        data = [d isa Number ? s*d : d for (s, d) in zip(scale, data[2:end])]
+        eventLibrary[id] = Dict("data"=>data)
     end
     eventLibrary
 end
```
cncastillo: Is this how Pulseq does it? Is the `format` kwarg there? Let's copy their logic.

I think a line like this is simpler to what is currently there:

`data = scale .* [data[2:end]...]'`

The data has a string or float, right? so multiplying by `""` is enough so the scaling works, no special case for not `Number`s is needed. At the `read_RF` level is gets converted to `char`?

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -410,22 +362,30 @@ function read_seq(filename)
                 end
                 blockEvents, blockDurations, delayInd_tmp = read_blocks(io, def["BlockDurationRaster"], pulseq_version)
             elseif  section == "[RF]"
-                if pulseq_version >= v"1.4.0"
+                if pulseq_version >= v"1.5.0"
+                    rfLibrary = read_events(io, [1/γ 1 1 1 1 1e-6 1 1 1 1 1]; format="%i "*"%f "^(10)*"%c ") # this is 1.5.x format
+                elseif pulseq_version >= v"1.4.0"
```
cncastillo, line 410: Check how Pulseq handles the change in format and this new char. I think they define an auxilary format function.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -410,22 +362,30 @@ function read_seq(filename)
...
-                if pulseq_version >= v"1.4.0"
-                    gradLibrary = read_events(io, [1/γ 1 1 1e-6]; type='g', eventLibrary=gradLibrary) # this is 1.4.x format
+                if pulseq_version >= v"1.5.0"
+                    gradLibrary = read_events(io, [1/γ 1/γ 1/γ 1 1 1e-6]; eventLibrary=gradLibrary) # this is 1.5.x format
+                elseif pulseq_version >= v"1.4.0"
+                    gradLibrary = read_events(io, [1/γ 1 1 1e-6]; eventLibrary=gradLibrary) # this is 1.4.x format
                 else
-                    gradLibrary = read_events(io, [1/γ 1 1e-6];   type='g', eventLibrary=gradLibrary) # this is 1.3.x and below
+                    gradLibrary = read_events(io, [1/γ 1 1e-6]; eventLibrary=gradLibrary) # this is 1.3.x and below
                 end
```
cncastillo, line 423: This `type` argument is how Pulseq handled the gradients, maybe they changed it, but if not, let's keep it.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -434,64 +394,61 @@ function read_seq(filename)
             elseif  section == "[SHAPES]"
                 shapeLibrary = read_shapes(io, (pulseq_version.major == 1 && pulseq_version.minor < 4))
             elseif  section == "[EXTENSIONS]"
-                extensionLibrary = read_extension_blocks(io)
+                extensionInstanceLibrary = read_events(io, [1 1 1]; format="%i "*"%i "^(3), eventLibrary=extensionInstanceLibrary)
```
cncastillo, line 440: Check how Pulseq handle the extensions.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -434,64 +394,61 @@ function read_seq(filename)
...
+                        extensionTypeLibrary[ext_id] = ext_type
+                        extensionSpecLibrary[ext_id] = read_events(io, KomaMRIBase.get_scale(ext_type); format="%i "*KomaMRIBase.get_scanf_format(ext_type))
                     else
-                        @warn "Ignoring unknown extension, input string: $extension"
+                        if ext_string in def["RequiredExtensions"]
+                            @error "Extension $ext_string is required by the sequence but not supported by KomaMRI reader"
+                        else
+                            @warn "Ignoring unsupported extension: $ext_string"
+                        end
                     end
```
cncastillo: I think the way extensions are being handles so you can define new ones for simulation is ok. 

Not really sure why `RequiredExtensions` is needed.

Type checks like `if ext_type <: Extension` are red flags, is that really needed? Potentially some of the logic inside the if can be simplified using multiple dispatch.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -434,64 +394,61 @@ function read_seq(filename)
...
                 else
                     @error "Unknown section code: $section"
                 end
             end
-
         end
     end
+    verify_signature!(filename, signature; pulseq_version=pulseq_version)
```
cncastillo, line 476:
```suggestion
    verify_signature!(filename, signature; pulseq_version)
```

Why do you need to verify the signature? Is that what Pulseq does?

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -434,64 +394,61 @@ function read_seq(filename)
...
     # fix blocks, gradients and RF objects imported from older versions
     if pulseq_version < v"1.4.0"
         # scan through the RF objects
-        for i = 0:length(rfLibrary)-1
-            rfLibrary[i]["data"] = [rfLibrary[i]["data"][1:3]' 0.0 rfLibrary[i]["data"][4:end]']
+        for i in eachindex(rfLibrary)
+            rfLibrary[i]["data"] = [rfLibrary[i]["data"][1:3]; 0.0; rfLibrary[i]["data"][4:end]]
         end
-        # scan through the gradient objects and update 't'-s (trapezoids) und 'g'-s (free-shape gradients)
```
cncastillo: Why is this change needed? That modifies the size of `rfLibrary[i]["data"]`.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -543,50 +501,63 @@ end
...
-            G = Grad(gA, gT, Δt_gr/2, Δt_gr/2, delay)
+        if time_shape_id == 0 #no time waveform. Default time raster
+            gT = Ngr * Δt_gr
+            G = Grad(gA, gT, Δt_gr/2, Δt_gr/2, delay, first_grads, last_grads)
+        elseif time_shape_id == -1 #New in pulseq 1.5.x: no time waveform. 1/2 of the default time raster
+            gT = Ngr * Δt_gr / 2
+            G = Grad(gA, gT, Δt_gr/2, Δt_gr/2, delay, first_grads, last_grads)
         else
             gt = decompress_shape(shapeLibrary[time_shape_id]...)
+            delay += gt[1] * Δt_gr # offset due to the shape starting at a non-zero value
```
cncastillo: are you sure this matches what Pulseq does? I am not sure if non-zero start values are allowed, the delay wouldn't make sense there.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -630,14 +616,21 @@ function read_RF(rfLibrary, shapeLibrary, Δt_rf, i)
...
     #Creating timings
-    if time_shape_id == 0 #no time waveform
+    if time_shape_id == 0 #no time waveform. Default time raster
         rfT = Nrf * Δt_rf
-    else
+    elseif time_shape_id == -1 #New in pulseq 1.5.x: no time waveform. 1/2 of the default time raster
+        rfT =  Nrf * Δt_rf / 2
+    else #time waveform
         rft = decompress_shape(shapeLibrary[time_shape_id]...)
+        delay += rft[1] * Δt_rf # offset due to the shape starting at a non-zero value
```
cncastillo, line 622: are you sure this matches what Pulseq does? I am not sure if non-zero start values are allowed, the delay wouldn't make sense there.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -659,17 +652,31 @@ function read_ADC(adcLibrary, i)
...
+    if length(a) == 8 # for version 1.5.x
+        #(1)num (2)dwell (3)delay (4)freq_ppm (5)phase_ppm (6)freq (7)phase (8)phase_shape_id
+        num       = a[1] |> x->floor(Int64,x)
+        dwell     = a[2]
+        delay     = a[3] + dwell/2
+        freq_ppm  = a[4]
+        phase_ppm = a[5]
+        freq      = a[6]
+        phase     = a[7]
+        phase_id  = a[8] |> x->floor(Int64,x)
```
cncastillo: Do you know why we would need `freq_ppm`, `phase_ppm` and `phase_shape_id` for the ADC? Just curious, we can implement them later.

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -686,13 +693,13 @@ Block sequence definition. Used internally by [`read_seq`](@ref).
 """
 function get_block(obj, i)
     #Unpacking
-    idelay, irf, ix, iy, iz, iadc, iext = obj["blockEvents"][i]
+    idur, irf, igx, igy, igz, iadc, iext = obj["blockEvents"][i]
```
cncastillo, line 683: Is `idur` not used?

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -718,53 +724,168 @@ function get_block(obj, i)
...
+        parsed_hash_norm = lowercase(replace(parsed_hash, " " => ""))
+        if parsed_hash_norm != lowercase(replace(sig_hash, " " => ""))
+            @warn "Pulseq signature Hash mismatch between parser and metadata. Using metadata value." filename
+        end
+    end
+    if computed_hash != expected_hash
+        @warn "Pulseq signature verification failed for $(basename(filename)). Expected $(expected_hash), computed $(computed_hash). The file may have been modified or generated with a different implementation." filename
+        # Don't error, just warn - the file can still be used
+    end
+end
```
cncastillo, line 872: Just quickly check that this is how Pulseq does it.

KomaMRIFiles/test/runtests.jl
```diff
@@ -8,25 +8,30 @@ using TestItems, TestItemRunner, KomaMRIBase
...
         @test seq.DEF["PulseqVersion"] == v"1.2.1"
-        @test seq.DEF["signature"] == "f291a24409c3e8de01ddb93e124d9ff2"
+        @test seq.DEF["signature"][:hash] == "f291a24409c3e8de01ddb93e124d9ff2"
 
-        seq = @suppress read_seq(pth*"/test_files/radial_JEMRIS.seq") #Pulseq v1.2.1
+        seq = @suppress read_seq(pth*"/test_files/pulseq/v1.2/radial_JEMRIS.seq") #Pulseq v1.2.1
         @test seq.DEF["FileName"] == "radial_JEMRIS.seq"
         @test seq.DEF["PulseqVersion"] == v"1.2.1"
-        @test seq.DEF["signature"] == "e827cfff4436b65a6341a4fa0f6deb07"
+        @test seq.DEF["signature"][:hash] == "e827cfff4436b65a6341a4fa0f6deb07"
```
cncastillo: I don't get why `signature` has a field `:hash`, is this consistent with Pulseq?

KomaMRIFiles/test/runtests.jl
```diff
@@ -118,21 +123,24 @@ end
...
-            for i in 1:length(seq_koma)
-                blk_koma   = get_samples(seq_koma, i)
-                blk_pulseq = NamedTuple{keys(blk_koma)}(seq_pulseq[i]) # Reorder keys
-                for (ev_koma, ev_pulseq) in Iterators.filter(not_empty, zip(blk_koma, blk_pulseq))
-                    @test ev_koma.t ≈ ev_pulseq.t
-                    @test inside(ev_koma.A) ≈ inside(ev_pulseq.A)
-                    @test first(ev_koma.A)  ≈ first(ev_pulseq.A) || ev_koma.t[2] ≈ ev_koma.t[1]
-                    @test last(ev_koma.A)   ≈ last(ev_pulseq.A)
+    pth          = joinpath(@__DIR__, "test_files/pulseq/pulseq_read_comparison/")
+    versions     = ["v1.4", "v1.5"]
```
cncastillo: No tests for versions lower than 1.4?

KomaMRIPlots/src/ui/DisplayFunctions.jl
cncastillo, line 1: It would very nice if a "x" or a "|" was displayed in `plot_seq` to check the RF center (like MATLAB's Pulseq does), as it becomes kind of critical to check the RF phase at the center (this will affect simulation).

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -509,11 +509,11 @@ function plot_M0(
     )
     p[4] = scatter(;
         x=t[rf_idx] * 1e3,
-        y=rf_type,
+        y=t[rf_idx] * 0,
         name="RFs",
```
cncastillo:
```suggestion
        name="RF_center",
```

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -598,11 +598,11 @@ function plot_M1(
...
```
cncastillo:
```suggestion
        name="RF_center",
```

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -686,11 +686,11 @@ function plot_M2(
...
```
cncastillo:
```suggestion
        name="RF_center",
```

KomaMRIBase/src/datatypes/Sequence.jl
cncastillo: Up to you if you want to implement it in this PR, the same logic applies to differentiate between the fields `A` and `T` being a number or a vector. As it requires a change in KomaMRIBase, we could put is as an issue for now (`RF` -> `RF{Amp, Time, RFUse}` for dispatching). I realized this was discussed already in an issue https://github.com/JuliaHealth/KomaMRI.jl/issues/362.

pvillacorta: How about this "temporary" solution? Same logic as with `center`: 

```julia
mutable struct RF
    A
    T
    Δf
    delay::Float64
    center::Float64
    use::RFUse
    RF(A, T, Δf, delay, center, use) = any(T .< 0) || delay < 0 ? error("RF timings must be non-negative.") : new(A, T, Δf, delay, center, use)
    RF(A, T, Δf, delay, center)      = _RF_with_center_and_use(A, T, Δf,  delay; center=center)
    RF(A, T, Δf, delay)              = _RF_with_center_and_use(A, T, Δf,  delay)
    RF(A, T, Δf)                     = _RF_with_center_and_use(A, T, Δf,  0.0)
    RF(A, T)                         = _RF_with_center_and_use(A, T, 0.0, 0.0)
end

function _RF_with_center_and_use(A, T, Δf, delay; center=0.0)
    rf = RF(A, T, Δf, delay, center, Undefined())
    rf.center = center == 0.0  ? get_RF_center(rf) : center
    rf.use = get_flip_angle(rf) <= 90.01 ? Excitation() : Refocusing()
    return rf
end
```

That does not mean we cannot implement `RF -> RF{Amp, Time, RFUse}` later (but yes, I think that we can put it as an issue for now)

KomaMRIFiles/test/runtests.jl
pvillacorta: The Pulseq's signature has 2 fields: "type" (the hashing algorithm type) and "hash" (the signature itself). So yes, this is consistent.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: This is done to check that no "manual" change has been made within the Pulseq file. The function throws a warning in case the read signature and the one computed from the payload no not match. 
That check is in fact, not conducted in the MATLAB API, [but in the C++](https://github.com/pulseq/pulseq/blob/28c55ea9050f93c28faa032185527459c29d69c8/src/ExternalSequence.cpp#L934-L970). Anyways, I think it is convenient to do it in Koma, just to know if something has been tuned outside the "oficial" Pulseq writers.

KomaMRIBase/src/datatypes/Sequence.jl
cncastillo: That seems ok, but I would change `center = 0.0` to `center = nothing`, so if someone actually want to have the center at 0 it doesn't generate problems.

Not sure why you defined `rf.use = Undefined()`

If you later change it to `rf.use = get_flip_angle(rf) <= 90.01 ? Excitation() : Refocusing()`

cncastillo: Oh it is to not use the constructor recursively.

KomaMRIBase/src/datatypes/Sequence.jl
pvillacorta: Done, but why is it a breaking change in KomaFiles?

KomaMRIBase/src/timing/TimeStepCalculation.jl
pvillacorta: You are absolutely right. To check it, I used Pulseq's MATLAB API to define an RF pulse with a duration of 10 ms and a delay of 0.7 ms, and checked that the delay is not included in the center calculation:
```
# Format of blocks:
# NUM DUR RF  GX  GY  GZ  ADC  EXT
[BLOCKS]
1 1070   1   0   0   0  0  0
2 1070   1   0   0   0  0  0
3 1070   1   0   0   0  0  0

# Format of RF events:
# id ampl. mag_id phase_id time_shape_id center delay freqPPM phasePPM freq phase use
# ..   Hz      ..       ..            ..     us    us     ppm  rad/MHz   Hz   rad  ..
# Field 'use' is the initial of: 
#   excitation refocusing inversion saturation preparation other undefined
[RF]
1           25 1 2 3 5000 700 0 0 0 0 e

# Sequence Shapes
[SHAPES]

shape_id 1
num_samples 2
1
1

shape_id 2
num_samples 2
0
0

shape_id 3
num_samples 2
0
10000
```

I think that, in order to stick to Pulseq, the most appropriate change is to make the `get_RF_center` function calculate the center without considering the delay. This way, the `center` field of the `RF` Koma struct would match with the Pulseq's center. Consequently, the line above should look like this:
```julia
rf0 = t0 + delay + center
```

KomaMRIBase/src/datatypes/Sequence.jl
cncastillo: Removing skip_rf could break someone's code (if they use it) as the function signature changes. But I'm pretty sure I'm the only person in the world using that feature. I'm fine removing it as a non breaking change.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: > Is this how Pulseq does it? Is the format kwarg there? Let's copy their logic.

See [the next comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/614#discussion_r2680448616)

> `data = scale .* [data[2:end]...]'`

There are two factors that prevent us from using the line above:
1. RF type is not a String but a Char. The `*` operator is not defined for chars
2. If we multiplied by "", `scale` would need to be of type `Any[]`. As it is defined now, `scale` type is `Float64[]`, which is more efficient

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: They have dedicated "libraries" (Dict) for each extension. This is not very extensible, that's why implemented these three dicitionaries:  `extensionInstanceLibrary`, `extensionTypeLibrary`, `extensionSpecLibrary`.

pvillacorta: Their  `extensionInstanceLibrary` is just called `extensionLibrary`. I can rename it if you consider it, but I think that the former is more clear.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: > Not really sure why `RequiredExtensions` is needed.

SInce Pulseq 1.5.1:
Screenshot 2026-01-09 at 13 16 12

That check is therefore to throw an error in case the unknown extension is inside the "Required Extensions".

> Type checks like if `ext_type <: Extension` are red flags, is that really needed? Potentially some of the logic inside the if can be simplified using multiple dispatch.

Solved (including the required extensions check) with method dispatch.
https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: Before the latest changes, some "empty" and unnecessary elements were added to `rfLibrary`, `gradLibrary`, etc (see changes in `read_events`).

Additionally, the "data" field of each one of these dictionaries is now a column vector and not a row vector. Therefore, there is no need to transpose.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: It was redundant, as the block's duration was already stored in the `blockDurations` dictionary. 
I solved this by removing `blockDurations` and `delayInd_tmp` and performing the version check inside `get_block`. That `idur` variable has different meaning depending on the Pulseq version:
- Version 1.4.0 or higher: Explicit block duration (in units of blockDurationRaster).
- Versions lower than 1.4: DELAYS table entry

Solved https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

KomaMRIFiles/test/runtests.jl
pvillacorta: I currently don’t have an easy way to generate test data for Pulseq < 1.4. Supporting it would require setting up older MATLAB versions and regenerating .seq and .mat files. This is doable, but I wasn’t sure whether <1.4 is still actively used (e.g., by JEMRIS). Happy to add this if it’s considered important.

KomaMRIBase/src/timing/TimeStepCalculation.jl
pvillacorta: Solved https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

KomaMRIBase/src/datatypes/Sequence.jl
pvillacorta: Solved https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

KomaMRIBase/src/datatypes/Sequence.jl
pvillacorta: Solved https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: Solved https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: > Put isnothing(signature_type) && isnothing(signature_hash) into a variable has_hash (?)

Solved https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/df79e7b3160b1050f9dcc7e2039d030b824d350a

> Why not return an empty string? what does Pulseq do?

I considered better to return `nothing` so we can do a better handling of the signature after reading it. Pulseq (remember, [the C++ API](https://github.com/pulseq/pulseq/blob/28c55ea9050f93c28faa032185527459c29d69c8/src/ExternalSequence.cpp#L934-L970)) has a boolean variable called `m_bSignatureDefined`. Although not the same, both ways to handle it are very similar (and I think ours is even a bit clearer).

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: See [this](https://github.com/pulseq/pulseq/blob/v1.5.1/matlab/%2Bmr/%40Sequence/read.m#L119-L120) and [this](https://github.com/pulseq/pulseq/blob/v1.5.1/matlab/%2Bmr/%40Sequence/read.m#L547-L557). This is also one of the things I think we do very similarly to Pulseq, but simpler. They use a NaN in the `scale` array and then detect that NaN in `readEvents` to define the format for the char. I think that a more direct (but similar) way to handle this is with the `format` kwarg.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: Pending to check

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: The other option is to completely mimic their approach (see [this comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/614#issuecomment-3372845782)).

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: The specification says nothing about time shapes compulsorily starting at zero. To verify for a specific case of non-zero starting time values, I have found that [writeSpiral.m (from Pulseq repository)](https://github.com/pulseq/pulseq/blob/v1.5.1/matlab/demoSeq/writeSpiral.m) generates a sequence whose first RF pulse has both:
- A delay of 100us:
    ```
    # Format of RF events:
    # id ampl. mag_id phase_id time_shape_id center delay freqPPM phasePPM freq phase use
    # ..   Hz      ..       ..            ..     us    us     ppm  rad/MHz   Hz   rad  ..
    # Field 'use' is the initial of: 
    #   excitation refocusing inversion saturation preparation other undefined
    [RF]
    1      125.953 1 2 3 4000 100 -3.35 0.0841947 0 0 s
    ```
- A time shape starting at a non-zero value (`5*rfRasterTime = 5us`):
    ```
    shape_id 3
    num_samples 800
    5
    10
    10
    797
    ```
These would give us a total delay of 105us. 
When plotting this sequence in MATLAB (even before exporting to .seq), we can see that this total delay is correctly applied:
Screenshot 2026-01-12 at 09 48 07

(when zoomed):
Screenshot 2026-01-12 at 09 43 21

pvillacorta: Another example is this MATLAB script:
```matlab
seq=mr.Sequence();   

sys = mr.opts('MaxGrad',20,'GradUnit','mT/m',...
    'MaxSlew',120,'SlewUnit','T/m/s');  

angles = linspace(-180, 180, 10);
shape_values = sinc(deg2rad(angles));

time_values = [2, 3, 4, 5, 7, 8, 10, 13, 16, 18] * 10 * sys.rfRasterTime;
rf = mr.makeArbitraryRf(shape_values, pi/1000, 'system', sys, 'delay', 300e-6, 'use','excitation');
rf.t = time_values;
rf.center = (time_values(end/2 + 1) + time_values(end/2))/2;
rf.shape_dur = time_values(end);

for i = 1:3
    seq.addBlock(rf);
end

seq.setDefinition('Name', 'rf-time-shaped');
seq.write('rf-time-shaped.seq') % Write to pulseq file  
```

Which generates a total delay of 0.3 + 0.02 = 0.32 ms. When plotting:
Screenshot 2026-01-12 at 10 17 15

pvillacorta: Not exactly the same happens with gradients (see [this comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/614#discussion_r2681507234))

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: (See [the rf-related comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/614#discussion_r2681322491) first)
This MATLAB script generates an arbitrary gradient with a 0.3 ms delay and a time shape starting at 0.2 ms:
```matlab
seq=mr.Sequence();   

sys = mr.opts('MaxGrad',20,'GradUnit','mT/m',...
    'MaxSlew',120,'SlewUnit','T/m/s',...
    'rfRingdownTime', 30e-6, 'rfDeadtime', 100e-6, 'gradRasterTime', 1e-5);  

amplitude = 42576; % Hz/m

angles = linspace(0, 360, 10);
shape_values = amplitude*cos(deg2rad(angles));

time_values = [2, 3, 4, 6, 7, 8, 10, 13, 16, 18] * 10 * sys.gradRasterTime;

sys = mr.opts('MaxGrad',20,'GradUnit','mT/m',...
    'MaxSlew',120,'SlewUnit','T/m/s',...
    'rfRingdownTime', 30e-6, 'rfDeadtime', 100e-6);  

gx = mr.makeArbitraryGrad('x', shape_values, sys, 'first', shape_values(1), 'last', shape_values(end), 'delay', 300e-6);
gx.tt = time_values;

for i = 1:3
    seq.addBlock(gx);
end

seq.setDefinition('Name', 'gr-time-shaped');
seq.write('gr-time-shaped.seq') % Write to pulseq file  
```
The total delay would be 0.5 ms, as reflected in the plot, however, also showing an initial sample at 0.3 ms:
Screenshot 2026-01-12 at 10 33 47

So, for the case of the gradients, it is not that clear that things are working well. I don't know which is the best solution for this.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: From the specification:
Screenshot 2026-01-12 at 10 56 10
So these are different ways to introduce a frequency/phase offset in the ADC. Not sure if they can be combined (eg. both `freq_ppm` and `freq` non-zero)

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: A quick comparison made with Cursor:

# Comparison with the Julia code

## Similarities
- Both read the hash in lowercase (Julia: `lowercase()`, C++: `str_tolower()`)
- Both ignore comments starting with `#`
- Both parse `Type` and `Hash` from the `[SIGNATURE]` section
- Both compare the stored hash with the computed one

## Important differences

### Algorithm support
- **Julia**: supports MD5, SHA1, and SHA256 (`supported_signature_digest`)
- **C++**: only supports MD5 (line 980: it only validates if it is `"md5"`)

### Handling of the newline before `[SIGNATURE]`
- **Julia (version ≥ 1.4.0)** tests three variants:
  - Exclude only the last newline (specification)
  - Include the last newline (some implementations such as MATLAB)
  - Exclude all consecutive newlines before `[SIGNATURE]`
  - Uses the first variant that matches
- **C++**: always includes the newline before `[SIGNATURE]` (fixed behavior)

### File reading
- **Julia**: reads the entire file as bytes and searches for `[SIGNATURE]` in the byte stream
- **C++**: processes the file line by line in raw mode while building the index

### Handling of spaces in the hash
- **Julia**: removes spaces from the hash using `replace(strip(value), " " => "")`
- **C++**: uses `str_trim()`, which only removes spaces at the beginning and end, not in the middle

### Pulseq version handling
- **Julia**: has conditional logic depending on `pulseq_version` (different behavior for `< 1.4.0` vs `≥ 1.4.0`)
- **C++**: fixed behavior without considering the version

## Conclusion

The C++ Pulseq code is simpler but less flexible:
- Only supports MD5
- Fixed newline handling (always included)
- No version awareness

The Julia code is more robust:
- Supports multiple algorithms
- Tries multiple newline variants
- Takes the Pulseq version into account
- More flexible handling of spaces in the hash

pvillacorta:
I made the last commit in order to (we can talk about it more in detail in a meeting if you consider):
- Simplify "one-value" waveforms before intantiating the actual KomaMRI events (`RF` or `Grad`)
    - I used a new function: `simplify_waveforms`
    - This made me find a discrepancy between our implementation of `fix_first_last_grads!` and [theirs](https://github.com/pulseq/pulseq/blob/v1.5.1/matlab/%2Bmr/%40Sequence/read.m#L325-L413): they update the gradLibrary dictionary with the obtained `first` and `last` values, and we were not doing that.  I have made the necessary changes to make them do the same. Again, we can better explore this if you need.
- Fix bugs in `plot_seq. RF were not displaying correctly when more than one RF pulses were active in the sequence.

cncastillo:
I think we might need to organize a meeting there's too many changes.

I don't understand why the `fix_first_last_grads!` needs to be applied to the dictionary. Is there a difference in the obtained `first` and `last` points? Before they didn't always match Pulseq's, now they always match?

Also I noticed that some of the tests use `inside(x)=x[2:end-1]` to compare so we don't consider the first and last element of the vector `A`, but now, `first` and `last` are in the `Grad` data structure. There seems to be a mismatch.

The functions `times` and `ampls` are to get samples for simulation and plotting. I don't really understand the addition of the centers, it doesn't seem to be the right place.

Can you explain the bug you encountered with `plot_seq`?

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: Maybe [this comment in a recent issue I posted in Pulseq's repo](https://github.com/pulseq/pulseq/issues/188#issuecomment-3541588756) can help us.
_"The first number stored in the time vector is de-facto 0 and the duration of the gradient is defined by the last time point"_
So non-zero start values (in gradients, RF still work) should not be accepted. I would throw a warning in case a time shape of a gradient starts at a non-zero value. However, I would allow reading the sequence anyways.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: I think that the best approach is as it is right now (`format` as a keyword argument and no `NaN` values within the `scale` array), even if it does not exactly mimic their MATLAB functions. This is because of two reasons:
- We use `read_events` also for reading extension specs. This is something that MATLAB API does not do, but that simplifies things a lot. Our only "drawback" is that not all numeric values read from an event are floats, some extensions have int values as well:

  ```julia
  mutable struct Trigger <: Extension 
      type::Int # Type of trigger (system dependent). 0: undefined / unused
      channel::Int # channel of trigger (system dependent). 0: undefined / unused
      d1::Float64 # Delay prior to the trigger event [s]
      d2::Float64 # Duration of trigger event [s]
  end
  ```
  
  That justifies the necessity of defining a dedicated `get_scanf_format` method for extensions.
- Additionally, while a (Char)*NaN method is defined in MATLAB:
  ```matlab-repl
  >> 'a' * NaN
  ans =
     NaN
  ```
  this is not the case in Julia:
  ```julia-repl
  julia> 'a'*NaN
  ERROR: MethodError: no method matching *(::Char, ::Float64)
  The function `*` exists, but no method is defined for this combination of argument types.
  
  Closest candidates are:
    *(::Any, ::Any, ::Any, ::Any...)
     @ Base operators.jl:642
    *(::Bool, ::T) where T<:AbstractFloat
     @ Base bool.jl:181
    *(::Missing, ::Number)
     @ Base missing.jl:123
    ...
  ```
  So things may be a bit more difficult to handle in this case.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: In yesterday's meeting we decided to do this:
> I would throw a warning in case a time shape of a gradient starts at a non-zero value. However, I would allow reading the sequence anyways.

pvillacorta:
This comment summarizes the points discussed in yesterday’s meeting with @cncastillo and their current status after the latest commit (https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/9c840750018e934e7d8d42c6de6f4a01f4a301ce). Some items required code changes (included in the commit), while others were already resolved and have now been verified:

- [x] **Simplify center plotting in `plot_seq`**  
  Addressed in the latest commit.
- [x] **`format_helper`**  
  Verified as discussed. See this [comment](https://github.com/JuliaHealth/KomaMRI.jl/pull/614#discussion_r2697616752) for confirmation and resolve it if you agree.
- [x] **Test `time_shape_id = -1`**  
  The `spiral.seq` file (v1.5) already contains gradients with `time_shape_id = -1`.  
  See the [file that generates this sequence](https://github.com/pulseq/pulseq/blob/v1.5.1/matlab/demoSeq/writeSpiral.m).
- [x] **Emit a warning if the gradient time shape does not start at 0**  
  Implemented in the latest commit.
- [x] **Review how Pulseq detects the version in `get_block`**  
  Pulseq does not have a `get_block` function, as this is specific to Koma. In Koma, `get_block` is used to convert the internal dictionaries (`obj.rfLibrary`, `obj.gradLibrary`, …) into `Sequence` instances. Pulseq itself stops at the previous step and defines the sequence directly using dictionaries stored in `obj`.

Additionally, I have changed the information message that appears when loading a v1.5 sequence:
```julia-repl
┌ Info: Pulseq 1.5.1 is supported, but Soft Delay, Rotation, and RF Shimming extensions are not yet included
└ (see https://github.com/JuliaHealth/KomaMRI.jl/issues/714)
```
The new issue #714 accounts for Pulseq v1.5 extension support.

KomaMRIFiles/src/Sequence/Pulseq.jl
cncastillo: Ok!

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -255,6 +259,8 @@ function plot_seq(
     rf = (
         A=reduce(vcat, [usrf(block.rf.A); Inf] for block in seq_samples),
         t=reduce(vcat, [usrf(block.rf.t); Inf] for block in seq_samples),
+        ct=center_times,
+        cA=abs.(KomaMRIBase.get_rfs(seq, center_times)[1])
```
cncastillo: Another important parameter to show at the RF center is the phase.

KomaMRIPlots/src/ui/DisplayFunctions.jl
```diff
@@ -357,11 +363,23 @@ function plot_seq(
...
+            y=rf.cA * 1e6 * frf,
+            name="RF_center",
+            hovertemplate="RF center: %{x:.4f} ms<br>Amplitude: %{y:.2f} μT<extra></extra>",
+            xaxis=xaxis,
+            yaxis=yaxis,
+            legendgroup="RF_center",
+            showlegend=showlegend,
+            mode="markers",
+            marker=attr(; color="#FF0000", symbol="x"),
+        )
```
cncastillo: For very long sequences (> 1000 RFs) can you check if this is too slow?, maybe the center display can be turned off by default.

KomaMRIFiles/test/runtests.jl
cncastillo: I think is ok to only support >=1.4

cncastillo: Actually at least 1.3 could be important for some people

https://github.com/JuliaHealth/KomaMRI.jl/issues/367

KomaMRIPlots/src/ui/DisplayFunctions.jl
pvillacorta: Do you mean the "accumulated" phase (as we talked last meeting) at that time point, or just `angle.(KomaMRIBase.get_rfs(seq, center_times)[1])`?

cncastillo: the second one! when writing the file, that phase will correspond to `phase_offset`.

KomaMRIPlots/src/ui/DisplayFunctions.jl
pvillacorta:
```julia-repl
julia> rf = PulseDesigner.RF_sinc(1e-5, 1e-3, Scanner())
Sequence[ τ = 1.5 ms | blocks: 2 | ADC: 0 | GR: 0 | RF: 1 | EXT: 0 | DEF: 0 ]

julia> seq = sum([rf for i in 1:1000])
Sequence[ τ = 1500.0 ms | blocks: 2000 | ADC: 0 | GR: 0 | RF: 1000 | EXT: 0 | DEF: 0 ]
```
Master branch:
```julia
julia> @time plot_seq(seq)
  8.478312 seconds (18.75 M allocations: 4.499 GiB, 6.44% gc time, 82.34% compilation time: <1% of which was recompilation)

julia> @time plot_seq(seq)
  1.009890 seconds (5.90 M allocations: 3.884 GiB, 10.37% gc time)
```

This branch:
```julia
julia> @time plot_seq(seq)
  9.331204 seconds (20.13 M allocations: 4.590 GiB, 6.28% gc time, 83.26% compilation time: <1% of which was recompilation)

julia> @time plot_seq(seq)
  1.070176 seconds (6.08 M allocations: 3.920 GiB, 9.92% gc time)
```

It is a bit slower, as expected. Anyway, I have turnt it off by default.

cncastillo: That's ok, feel free to leave it on by default!

cncastillo: That's ok, feel free to leave it on by default!

cncastillo: That's ok, feel free to leave it on by default!

cncastillo: That's ok, feel free to leave it on by default!

pvillacorta: See https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/eaf7f18a9b139fe074cfb636d029b48ac422b527.

I disabled this trace by default. Based on your comment, it felt better to only enable it when needed, so that the default plot remains clear :)

KomaMRIPlots/src/ui/DisplayFunctions.jl
pvillacorta: See https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/eaf7f18a9b139fe074cfb636d029b48ac422b527

KomaMRIFiles/test/runtests.jl
pvillacorta: See https://github.com/JuliaHealth/KomaMRI.jl/pull/614/commits/eaf7f18a9b139fe074cfb636d029b48ac422b527.

I had to solve some timing-related problems for <v1.4, since I thought that the delay of each block was added to the duration of its events, but [they actually use that delay value as a kind of "duration"](https://github.com/pulseq/pulseq/blob/337e0936eba019cd6aa48d1a8ee6bd8d8bb5f094/matlab/%2Bmr/calcDuration.m#L24-L25).

I also found that duration for blocks containing a uniformly-shaped RF was not correctly calculated, as it was lacking a final `Δt_rf/2` (see [slides 4 and 9](https://github.com/pulseq/pulseq/blob/master/doc/pulseq_shapes_and_times.pdf)).

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -44,141 +43,125 @@ function read_definitions(io)
...
-                eventTable[blockEvents[1]] = Int64[0; blockEvents[3:end]...]
-            end
-
-            if pulseq_version >= v"1.4.0"
-                blockDurations[blockEvents[1]] = blockEvents[2]*blockDurationRaster
-            else
-                delayIDs_tmp[blockEvents[1]] = blockEvents[2]
+                eventTable[blockEvents[1]] = Int64[blockEvents[2:end]...]
             end
         end
```
cncastillo, line 97: Doesn't this mess up <v1.2.1?

KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -434,98 +437,46 @@ function read_seq(filename)
...
-                gradLibrary[i]["data"] = [gradLibrary[i]["data"][1:2]; 0.0; gradLibrary[i]["data"][3:end]]
-            end
-        end
-        # for versions prior to 1.4.0 blockDurations have not been initialized
-        blockDurations = zeros(length(blockEvents))
-        for i = 1:length(blockEvents)
-        idelay = delayInd_tmp[i]
-            if idelay > 0
-                delay = tmp_delayLibrary[idelay]["data"][1]
-                blockDurations[i] = delay
```
cncastillo, line 501: Isn't this required for reading <v1.4.1? I think i copied this from Pulseq's reader.

pvillacorta: No, this is no longer needed, since all this version handling is now conducted within `get_RF` and `get_Grad`. Once again, we are not exactly replicating Pulseq, but I think it is worth it to change things a bit to save a few lines :)

Regarding lines 475-489, they ensured that zero-amplitude trapezoidal gradients had non-zero rise and delay when there was flat time, redistributing time from the flat to maintain the total duration. As I have tested, they seem to be no longer necessary.

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: No, as we now include block durations(v1.4+)/delay IDs(v1.3-) inside `blockEvents` dictionary. 
Both `blockDurations` and `delayIDs_tmp` were removed for simplicity.

KomaMRIFiles/src/Sequence/Pulseq.jl
cncastillo: Can you point me to where the subtraction by the gradient raster is done in the new code? I think this mattered for pulseq 1.2 (the jemris exported pulseq files). 

If the tests are passing it should be ok tho.

aTrotier:
Nice 👍 I will give it a try !

PR #623, opened 2025-10-08
Support fully asymmetric `TimeCurve`s with `Periodic` constructor

### LINKED ISSUE #619: Fully asymmetric periodic timing
Hi Developers,

I am working with arbitrary motion (flowpath), and the paths are periodic, in that they should 'restart' once they reach the end of the definition.
I would like to use `Periodic` to achieve this, ideally with `asymmetry=1.0`, however this produces an error.
A workaround is to use something like `asymmetry=0.999`, but that feels a little hacky.
Another possible solution would be to add a "periodic" flag to a `TimeRange`.

### LINKED ISSUE #626: Problems with using `Periodic` in Julia 1.12
### What happened?

```julia
┌ KomaMRIBase
│  WARNING: Constructor for type "Periodic" was extended in `KomaMRIBase` without explicit qualification or import.
│    NOTE: Assumed "Periodic" refers to `Interpolations.Periodic`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Periodic end`.
│    Hint: To silence the warning, qualify `Periodic` as `Interpolations.Periodic` in the method signature or explicitly `import Interpolations: Periodic`.
└
```

### Environment

```Markdown
Julia 1.12
```

pvillacorta:
This PR addresses #619 

Closes #619
Closes #626

### DIFF
--- a/KomaMRIBase/src/motion/TimeCurve.jl
+++ b/KomaMRIBase/src/motion/TimeCurve.jl
@@ -114,9 +114,9 @@ julia> periodic = Periodic(period=1.0, asymmetry=0.2)
 ![Periodic](../assets/periodic.svg)
 """
 function Periodic(period::T, asymmetry::T) where T 
-    if asymmetry == 1
+    if asymmetry == oneunit(T)
         return TimeCurve(t=[zero(T), period], t_unit=[zero(T), oneunit(T)], periodic=true)
-    elseif asymmetry == 0
+    elseif asymmetry == zero(T)
         return TimeCurve(t=[zero(T), period], t_unit=[oneunit(T), zero(T)], periodic=true)
     else
         return TimeCurve(t=[zero(T), period*asymmetry, period], t_unit=[zero(T), oneunit(T), zero(T)], periodic=true)


pvillacorta:
I am investigating CI errors with julia 1.12

KomaMRIBase/src/motion/TimeCurve.jl
```diff
@@ -109,8 +113,16 @@ julia> periodic = Periodic(period=1.0, asymmetry=0.2)
 ```
 ![Periodic](../assets/periodic.svg)
 """
-Periodic(period::T, asymmetry::T) where T = TimeCurve(t=[zero(T), period*asymmetry, period], t_unit=[zero(T), oneunit(T), zero(T)], periodic=true)
-Periodic(; period=1.0, asymmetry=0.5)     = Periodic(period, asymmetry)
+function Periodic(period::T, asymmetry::T) where T 
+    if asymmetry == 1
+        return TimeCurve(t=[zero(T), period], t_unit=[zero(T), oneunit(T)], periodic=true)
+    elseif asymmetry == 0
```
cncastillo:
```suggestion
    if asymmetry == oneunit(T)
        return TimeCurve(t=[zero(T), period], t_unit=[zero(T), oneunit(T)], periodic=true)
    elseif asymmetry == zero(T)
```

Issue #620, opened 2025-10-06
Magnetization issues for long sequences in flow phantoms


cncastillo:
### Feature Request

After talking to @mcgrathcm, we realized it is not clear what happens for `FlowPath` motions when the `periodic` argument in `TimeCurve` is `true`.

We talked with @pvillacorta, and we realized that some problems can be solved by adding more "key points" to the simulation, to consider the the periodic case.

[2025-10-06-Note-09-39.pdf](https://github.com/user-attachments/files/22726933/2025-10-06-Note-09-39.pdf)

Questions for @pvillacorta:
- What additional points are we adding to the simulation for `Path`, and `FlowPath`?

Related #516

mcgrathcm:
For me it was generally unclear what occurs at the end of FlowPath, specifically in the case were the defined motion is not truly periodic, like for flow.

Since you are not enforcing any periodicity on the FlowPath itself, what occurs at the end seems ambiguous. Do the particles reappear at their starting point? Is their magnetization preserved or reset? Is there some smart mapping to nearest starting point? If so, what happens if there are spatial particle density changes that might interfere with a nice mapping?

Ideally, for flow simulation, there should be a step that remaps magnetization, such that if I took a snapshot of all particle before and after the end of the period, the overall magnetization distribution should looks the same, but the actual particle positions should reflect either the last or the first timepoint. 

Finally, I think it is especially unclear for users who have not familiarized themselves with the phantom format, as they might assume that everything is handled under the hood, and might misconstrue results from the periodic resetting with real MR artifacts.

PR #636, opened 2025-10-16
Fix bug in `times` function when `periods` is not a scalar

pvillacorta:
While addressing issue #620 (I haven’t yet submitted any PRs related to it), I discovered a bug in the times function located in KomaMRIBase/src/motion/TimeCurve.jl. The function wasn’t behaving as intended and only worked correctly when the last element of the argument t was equal to 1. (This happens to be the case in [this example](https://juliahealth.org/KomaMRI.jl/stable/tutorial/07-RRVariability/), which is why I hadn’t detected the bug until now). This PR addresses the problem. 

It might be a good idea to register a new version of KomaMRIBase after merging this. Alternatively, we could wait until issue #620 is resolved and possibly include a few more updates to the base. Although this PR fixes a bug, it affects a relatively uncommon use case (used only for arrhythmia emulation), so releasing a new version is not urgent.

### DIFF
--- a/KomaMRIBase/src/motion/TimeCurve.jl
+++ b/KomaMRIBase/src/motion/TimeCurve.jl
@@ -136,11 +136,9 @@ function times(t, per::Real)
     return per .* t
 end
 function times(t, per::AbstractVector)
-    tr      = repeat(t, length(per))
     scale   = repeat(per, inner=[length(t)])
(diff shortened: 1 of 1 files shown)


cncastillo:
I will just merge this and register a new version of KomaMRIBase

Issue #604, opened 2025-08-29
Optimize Rotation motion


pvillacorta:
### Feature Request

For now, simulating a `Rotate` action is much lower than including the same rotation into a `Path` action.
The `displacement!` functions for `Rotate` arguments must be optimized.

cncastillo:
Is this only for rotational motion? It would be great to have an idea on how slow these motions are compared to no motion:
- Translate
- Rotate
- Path
- HeartBeat
- FlowPath
- ...

In my tests Rotate was around 10x slower than no motion.

PR #638, opened 2025-10-20
Add key time points in `DiscreteSequence` when evaluating periodic and flow-related motions

pvillacorta:
Try with:
```julia
add https://github.com/JuliaHealth/KomaMRI.jl#key-time-points:KomaMRIBase
add https://github.com/JuliaHealth/KomaMRI.jl#key-time-points:KomaMRICore
add https://github.com/JuliaHealth/KomaMRI.jl#key-time-points:KomaMRIFiles
```
From https://github.com/JuliaHealth/KomaMRI.jl/issues/620#issuecomment-3380802095:

> We have the following implementation of the `add_jump_times!` function for `FlowPath` actions:
> 
> ``` julia
> function add_jump_times!(t, a::FlowPath, tc::TimeCurve) 
>     jump_times = (tc.t_end - tc.t_start)/(size(a.spin_reset)[2]-1) * (getindex.(findall(a.spin_reset .== 1), 2) .- 1) .- 1e-6
>     append!(t, jump_times)
> end
> ```
> which adds to the simulation the time points at which a spin reset is produced. The substraction of 1us (`.- 1e-6`) to these time points ensures that we are evaluating within the time interval in which interpolation will return 1 (related [#511](https://github.com/JuliaHealth/KomaMRI.jl/pull/511)). This way, we do not miss any spin reset.
> 
> However, this `add_jump_times!` function certainly presents some problems. I will put them as checkboxes so I can address them in a PR:
> 
> - [x]  Until now, I hadn't realized that, if `periodic == true`, no time points are added for next periods. I must extend the function with `push!(jump_times, jump_times .+ n*period)` or something similar.
> - [x]  Since a big "jump" or spatial discontinuity may be produced when spins re-appear at their starting point at the beginning of each period, we might well ensure that no phase is accumulated during a simulation time step (`Δt` in the drawing) due to this position change. To achieve this, we must include in `jump_times` two more time points (as depicted in blue in the drawing): `tf` and `tf+eps`.

This is the PR intended to solve the above. I have created and renamed some functions to obtain the following structure:
- `add_key_time_points!`: this function is called from `get_variable_times` and adds all the necessary time points to the sequence  when motion is present. These key points can be divided into two categories: period-related and flow-related, which are obtained through two different and consecutive function calls from `add_key_time_points!`:
  - `add_period_times!`: adds to the time vector all the key points related to pseudo-periodic motions (i.e., motions where the `periods` field is a vector). This is necessary for all motion types, not only `FlowPath` actions.
  - `add_reset_times!`: this function is equivalent to the old `add_jump_times!`, but it has been improved to support pseudo-periodicity.
  
  Finally, when `periodic == true`, all key time points obtained from both `add_period_times!` and `add_reset_times!` are periodically extended.

This PR does not fully resolve issue #620 but provides a partial fix. Based on the figure in [this comment](https://github.com/JuliaHealth/KomaMRI.jl/issues/620#issue-3488165675), it addresses the point marked with the blue bullet.

This change might slightly impact performance in dynamic simulations, since we now check periodicity and add key time points for all motion types, not just for FlowPath motions. I realized that any motion (not only FlowPath) can lead to discontinuities when extended periodically, which is why these checks are now always performed (except, of course, in the `NoMotion` case).

### DIFF
--- a/KomaMRIPlots/Project.toml
+++ b/KomaMRIPlots/Project.toml
@@ -23,7 +23,7 @@ KomaPlotsPlutoPlotlyExt = "PlutoPlotly"
 
 [compat]
 Interpolations = "0.13, 0.14, 0.15, 0.16"
-KomaMRIBase = "0.9.7"
+KomaMRIBase = "0.9.4"
 MAT = "0.10, 0.11"
 PlotlyJS = "0.18"
 PlutoPlotly = "0.4, 0.5, 0.6"


pvillacorta:
The way I tested the included functions is by manually determining key points from a specified "TimeCurve":

```julia
t        = [0.0, 0.1, 0.3]
t_unit   = [0.0, 0.4, 1.0]
periods  = [1.0, 0.5, 2.0]
periodic = true
dx = dy  = [0.0 0.0 0.0 0.0]
dz       = [3.0 4.0 -4. -3.]
reset    = [false false true false]
Δ        = 1e-6

fpth = flowpath(dx, dy, dz, reset, TimeCurve(t, t_unit, periodic, periods), AllSpins())
```

IMG_0840

```julia
period_times = reduce(vcat, [t .+ [-Δ, Δ] for t in [0.0, 0.3, 0.45, 1.05, 1.35, 1.5]])
reset_times  = [0.2, 0.4, 0.85, 1.25, 1.45] .- Δ
```

KomaMRIBase/src/motion/Action.jl
```diff
@@ -1,7 +1,7 @@
 abstract type AbstractAction{T<:Real} end
 
+Base.:(≈)(a, b) = (a isa Nothing || b isa Nothing) ? false : Base.≈(a, b)
 Base.:(≈)(::Nothing, ::Nothing) = true
```
cncastillo: This is type piracy. Only create new methods for your types. Please check that this does not happen elsewhere.

```julia
Base.:(≈)(a, b) = (a isa Nothing || b isa Nothing) ? false : Base.≈(a, b)
Base.:(≈)(::Nothing, ::Nothing) = true
```

KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -72,6 +72,9 @@ function Base.:(≈)(obj1::Phantom, obj2::Phantom)
     if length(obj1) != length(obj2) return false end
     return reduce(&, [getfield(obj1, field)  ≈ getfield(obj2, field) for field in NON_STRING_PHANTOM_FIELDS])
 end
+# Comparison between two different motion types is always false:
+Base.:(==)(::Union{NoMotion, Motion, MotionList}, ::Union{NoMotion, Motion, MotionList}) = false
+Base.:(≈)(::Union{NoMotion, Motion, MotionList},  ::Union{NoMotion, Motion, MotionList}) = false
```
cncastillo, line 77: Doesn't this also affect when both motions are of the same type?

KomaMRICore/test/runtests.jl
```diff
@@ -449,7 +449,7 @@ end
         rotate(0.0, 0.0, 45.0, TimeRange(0.0, 1.0)),
         heartbeat(-0.6, 0.0, 0.0, Periodic(period=1.0)),
         path([0.0 0.0], [0.0 1.0], [0.0 0.0], TimeRange(0.0, 10.0)),
-        flowpath([0.0 0.0], [0.0 1.0], [0.0 0.0], [false false], TimeRange(0.0, 10.0))
+        flowpath([0.0 0.0], [0.0 1.0], [0.0 0.0], [false false], TimeRange(0.0, 10.0)) # We should find a way to test this when spin_reset flags are true
```
cncastillo, line 452: Agreed.

KomaMRIBase/test/runtests.jl
```diff
@@ -647,6 +649,39 @@ end
...
+    @testset "Key Time Points" begin
+        # Sequence duration
+        t_start = 0.0
+        t_end   = 1.5 
+
+        # TimeCurve parameters
+        t        = [0.0, 0.1, 0.3]
+        t_unit   = [0.0, 0.4, 1.0]
+        periods  = [1.0, 0.5, 2.0]
+        periodic = true
```
cncastillo: Is `periodic = false` tested?

KomaMRIBase/test/runtests.jl
```diff
@@ -647,6 +649,39 @@ end
...
+        dx = dy = [0.0 0.0 0.0 0.0]
+        dz      = [3.0 4.0 -4. -3.]
+        reset   = [false false true false]
+        Δ = 1e-6
```
cncastillo: What is `Δ`? use a better name. If this is the "small time" used to reset, it needs to use be a constant that is **way smaller**, like MIN_RISE_TIME.

KomaMRIBase/test/runtests.jl
```diff
@@ -647,6 +649,39 @@ end
...
+
+        # Key time points ("manually" determined)
+        period_times = reduce(vcat, [t .+ [-Δ, Δ] for t in [0.0, 0.3, 0.45, 1.05, 1.35, 1.5]])
+        reset_times  = [0.2, 0.4, 0.85, 1.25, 1.45] .- Δ
+
+        # Any motion with no spin resets (only key time points derived from periods)
+        pth  = path(dx, dy, dz, TimeCurve(t, t_unit, periodic, periods), AllSpins())
+        tq = [t_start, t_end]
```
cncastillo: what is `tq` use a better name

KomaMRIBase/test/runtests.jl
```diff
@@ -647,6 +649,39 @@ end
...
+        KomaMRIBase.add_key_time_points!(tq, Δ, pth)
+        key_points = period_times
+        @test sort(tq[3:end]) ≈ key_points[t_start .< key_points .< t_end]
```
cncastillo: I don't understand what you are doing here.

KomaMRIBase/test/runtests.jl
```diff
@@ -647,6 +649,39 @@ end
...
+
+        # FlowPath with a spin reset (key time points derived from both periods and spin resets)
+        fpth = flowpath(dx, dy, dz, reset, TimeCurve(t, t_unit, periodic, periods), AllSpins())
+        tq = [t_start, t_end]
+        KomaMRIBase.add_key_time_points!(tq, Δ, fpth)
+        key_points = sort(vcat(period_times, reset_times))
+        @test sort(tq[3:end]) ≈ key_points[t_start .< key_points .< t_end]
```
cncastillo: I don't understand what you are doing here.

cncastillo (changes requested):

KomaMRIBase/src/datatypes/simulation/DiscreteSequence.jl
```diff
@@ -86,7 +86,7 @@ based on simulation parameters.
 - `seqd`: (`::DiscreteSequence`) DiscreteSequence struct
 """
 function discretize(seq::Sequence; sampling_params=default_sampling_params(), motion=NoMotion())
-    t, Δt      = get_variable_times(seq; Δt=sampling_params["Δt"], Δt_rf=sampling_params["Δt_rf"], motion=motion)
+    t, Δt      = get_variable_times(seq; Δt=sampling_params["Δt"], Δt_rf=sampling_params["Δt_rf"], Δt_rise=sampling_params["Δt_rise"], motion=motion)
```
cncastillo: don't introduce new variables like one.

KomaMRIBase/src/datatypes/Phantom.jl
pvillacorta: No, since this method has less type restrictions than the rest of the specific `==` and `≈` methods.

KomaMRIBase/test/runtests.jl
pvillacorta: It was thought as "query times", i.e., the vector of time instants which are introduced in the function and in which key points are added.
These would be, in fact, the times of the sequence (`seqd.t`), so I can call it `seqd_t` or something similar.

KomaMRIBase/test/runtests.jl
pvillacorta: Not explicitly, but `periodic = false` covers the same logic that the periodic case, except it skips the periodic extension (the `if periodic`), so it's implicitly covered. 
I can add a dedicated test if you prefer.

KomaMRIBase/test/runtests.jl
pvillacorta: I am recreating what is done in `get_variable_times`, assuming that the sequence time points are only two: `[t_start, t_end]`. Function `add_key_points!` appends the key time points to this sequence time vector (before called `tq`, now called `seqd_t`). For the non-reset case, only period-related times are added. Then I have to compare the added times with the vector manually determined.

KomaMRIBase/test/runtests.jl
pvillacorta: The same that above, but in this case there are spin resets. Key points are thus composed by both the period-related and the reset time points.

KomaMRIBase/src/datatypes/simulation/DiscreteSequence.jl
pvillacorta: Solved in 01390d4201a11dec1c318e0bb1b958e3f90ad9f4

KomaMRIBase/src/motion/Action.jl
pvillacorta: Solved in 01390d4201a11dec1c318e0bb1b958e3f90ad9f4

KomaMRIBase/test/runtests.jl
pvillacorta: Solved in 01390d4201a11dec1c318e0bb1b958e3f90ad9f4

KomaMRIBase/test/runtests.jl
pvillacorta: I tried to clarify this a bit on 01390d4201a11dec1c318e0bb1b958e3f90ad9f4.
I also added a test for the MotionList case.

KomaMRIBase/test/runtests.jl
pvillacorta: Clarified in 01390d4201a11dec1c318e0bb1b958e3f90ad9f4.

KomaMRIBase/test/runtests.jl
pvillacorta: Renamed it to `seqd_t` in 01390d4201a11dec1c318e0bb1b958e3f90ad9f4

KomaMRIBase/src/motion/actions/simpleactions/Rotate.jl
```diff
@@ -1,5 +1,9 @@
+struct CenterOfMass end
+Base.:(≈)(::CenterOfMass, ::CenterOfMass) = true
+Base.:(≈)(a, b) = (a isa CenterOfMass || b isa CenterOfMass) ? false : Base.≈(a, b)
```
cncastillo: This is still type piracy. You are redefining `Base.:(≈)(a, b)`.

You probably want these:
```julia
Base.:(≈)(a::CenterOfMass, b)
Base.:(≈)(a, b::CenterOfMass)
```

KomaMRIBase/src/motion/actions/simpleactions/Rotate.jl
```diff
@@ -63,7 +67,7 @@ julia> r = Rotate(pitch=0.0, roll=45.0, yaw=0.0, center=(5e-3,0.0,0.0))
     pitch      :: T
     roll       :: T
     yaw        :: T
-    center     :: Union{Nothing,NTuple{3,T}} = nothing
+    center     :: Union{CenterOfMass,NTuple{3,T}} = CenterOfMass()
 end
 
 RotateX(pitch::T) where {T<:Real} = Rotate(pitch, zero(T), zero(T))
```
cncastillo: Can you quickly check that these are type stable:
- `displacement_x!(ux, action::Rotate, x, y, z, t)`
- `displacement_y!(ux, action::Rotate, x, y, z, t)`
- `displacement_z!(ux, action::Rotate, x, y, z, t)`
- `RotateX`, `RotateY`, `RotateZ`

using `@code_warntype` and `@device_code_warntype`.

KomaMRIBase/src/motion/TimeCurve.jl
```diff
@@ -129,20 +129,19 @@ Base.:(==)(t1::TimeCurve, t2::TimeCurve) = reduce(&, [getfield(t1, field) == get
...
-    offsets = repeat(cumsum(vcat(0, per[1:end-1]*t[end])), inner=[length(t)])
-    return (repeat(t, length(per)) .* scale) .+ offsets
-end
-function unit_time(tq, t, t_unit, periodic, per::Real)
-    return interpolate_times(t .* per, t_unit, periodic, tq)
-end
-function unit_time(tq, t, t_unit, periodic, per::AbstractVector)
-    return interpolate_times(times(t, per), repeat(t_unit, length(per)), periodic, tq)
+    offsets = repeat(cumsum(vcat(0, per[1:end-1]*(t_end - t_start))), inner=[length(t)])
+    return t_start .+ ((repeat(t, length(per)) .- t_start).* scale) .+ offsets
```
cncastillo: Why abbreviate, just use `periods` (?)

KomaMRIBase/test/runtests.jl
cncastillo: Please add one test or that case as well.

KomaMRIBase/test/runtests.jl
```diff
@@ -647,6 +646,38 @@ end
...
+        reset   = [false false true false]
+        ϵ = KomaMRIBase.MIN_RISE_TIME
+        # Key time points ("manually" determined):
+        period_times = [t+δ for t in (0.0, 0.3, 0.45, 1.05, 1.35, 1.5) for δ in (-ϵ, ϵ) if (t+δ) > t_start && (t+δ) < t_end]
+        reset_times  = [0.2, 0.4, 0.85, 1.25, 1.45] .- ϵ
+        # Any motion with no spin resets (only key time points derived from periods):
+        pth  = path(dx, dy, dz, TimeCurve(t, t_unit, periodic, periods), AllSpins())
+        seqd_t = [t_start, t_end]
+        KomaMRIBase.add_key_time_points!(seqd_t, pth)
+        @test sort(seqd_t[3:end]) ≈ period_times
```
cncastillo: This is much better, but why do you need to start comparing from 3? 

I think it is better to add the missing points at the right had side to make it more explicit `[-ϵ; 0; period_times]`, or something like that.

KomaMRIBase/src/motion/actions/simpleactions/Rotate.jl
pvillacorta: For `RotateX`:

image


For `displacement_x!`, `displacement_y!`, and `displacement_z!`, the only stability'related warnings are due to the `Union{CenterOfMass,NTuple}` in the rotation center type. This is  not even an error, as it is yellow, and not red. Anyway, I’ve simplified these displacement functions and created [`get_center`](https://github.com/JuliaHealth/KomaMRI.jl/pull/638/commits/8b80c4fac6b0c8f971fb0e21f9d2d808abf2523a#diff-4585bb242fc1c6caa26d49438146481c731c22677e6dac5d99d51d8a5fd4350f), which determines the rotation center via dispatch. This smooths things out a bit:

image
image
image

If we wanted to get rid of that single warning, we’d need to create an abstract type called `RotationCenter` or something similar. Personally, I don’t think it’s worth it — at least not until we actually optimize these functions (see #604).

Here’s the log from `@device_code_warntype KomaMRIBase.displacement_x!(ux_d, rt_d, x_d, y_d, z_d, t_d)` (only `Union`s and `dynamic` tags appear as warnings): [log.txt](https://github.com/user-attachments/files/23416536/log.txt)

KomaMRIBase/test/runtests.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/638/commits/8b80c4fac6b0c8f971fb0e21f9d2d808abf2523a

KomaMRIBase/src/motion/TimeCurve.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/638/commits/8b80c4fac6b0c8f971fb0e21f9d2d808abf2523a

KomaMRIBase/src/motion/actions/simpleactions/Rotate.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/638/commits/8b80c4fac6b0c8f971fb0e21f9d2d808abf2523a

KomaMRIBase/test/runtests.jl
pvillacorta: Solved in https://github.com/JuliaHealth/KomaMRI.jl/pull/638/commits/8b80c4fac6b0c8f971fb0e21f9d2d808abf2523a

KomaMRIBase/src/motion/actions/simpleactions/Rotate.jl
cncastillo: The warning is ok, is the union of two things, that should be ok.

cncastillo:
Re-running again the failed job, let me know if it passes to bump and register after merge.

cncastillo:
That is weird, I run it again with debug logs on, to see what is going on.

pvillacorta:
I still don't get why docs are failing :-/

cncastillo:
I can't figure out it maybe add a timeout, like this one:

https://github.com/JuliaHealth/KomaMRI.jl/blob/master/.github%2Fworkflows%2FCI.yml#L15

KomaMRIBase/src/datatypes/Phantom.jl
cncastillo: This is not true, it does affect the case where the types are the same. MWE:

```julia
struct Test1
    x
end

struct Test2
    x
end

f(x::Union{Test1, Test2}, y::Union{Test1, Test2}) = println("hello")

julia> f(Test1(1.0), Test2(2.0))
hello

julia> f(Test1(1.0), Test1(2.0))
hello
```

Now, i see what you mean, that there might be one more specific one like `f(x::Test1, y::Test1) = println("hello2")`, and therefore it doesnt affect in this case.

pvillacorta:
It is #620

pvillacorta:
Very nice example and simulations :) And thank you for testing this!
I think that, even merging this PR, we might not want to close #620 yet, since there is still a [pending change](https://github.com/JuliaHealth/KomaMRI.jl/issues/620#issuecomment-3380802095) for the future:

> Ideally, for flow simulation, there should be a step that remaps magnetization, such that if I took a snapshot of all particle before and after the end of the period, the overall magnetization distribution should looks the same, but the actual particle positions should reflect either the last or the first timepoint.

(PD: I have just noticed that these test images are gifs :O)

KomaMRICore/Project.toml
cncastillo:
```suggestion
KomaMRIBase = "0.9.7"
```

KomaMRIFiles/Project.toml
cncastillo:
```suggestion
KomaMRIBase = "0.9.7"
```

.github/workflows/CI.yml
```diff
@@ -156,6 +156,7 @@ jobs:
     if: '!cancelled()'
     name: Documentation
     runs-on: ubuntu-latest
+    timeout-minutes: 60
```
cncastillo: Remove this line.

cncastillo (approved):
Just check that the versions of the files are above master and the compats make sense as I think I registered a minor version for komafiles.

.github/workflows/CI.yml
pvillacorta: Solved in [6c15d00](https://github.com/JuliaHealth/KomaMRI.jl/pull/638/commits/6c15d0076b7989976533a4b1f78f1910e815d0d0)

pvillacorta:
> Just check that the versions of the files are above master and the compats make sense as I think I registered a minor version for komafiles.

KomaFiles version is set to 0.9.7 in master. This PR sets it to 0.9.8

KomaMRIPlots/Project.toml
```diff
@@ -23,7 +23,7 @@ KomaPlotsPlutoPlotlyExt = "PlutoPlotly"
 
 [compat]
 Interpolations = "0.13, 0.14, 0.15, 0.16"
-KomaMRIBase = "0.9.4"
+KomaMRIBase = "0.9.7"
```
cncastillo: No need to change this compat, as you didnt change KomaMRIPlots.

```suggestion
KomaMRIBase = "0.9.4"
```

PR #643, opened 2025-11-14
Higher order gradients

mcgrathcm:
Introduce new typing for discrete sequences and enhance the simulation core to support higher order gradients, based on spatial spherical harmonics.


mcgrathcm:
Closed to revisit after some other changes.

Issue #644, opened 2025-11-17
Reconstruction dimensions are not read correctly from Pulseq definitions


thePlasm:
### What happened?

The current code that reads the `Nx`, `Ny` and `Nz` definitions from Pulseq files reads them as floating point numbers instead of integers. This results in a `MethodError` when performing reconstruction as one of the reconstruction functions expects the reconstruction dimensions to be integers. This is especially important for testing non-cartesian sequences as the default values for `Nx` and `Ny` from ADC events do not necessarily apply.

As an example, if I use the `spiral.seq` file in the `examples/1.sequences` directory with `Nx 100`, `Ny 100` and `Nz 1` added to the definitions, the reconstruction fails with the stacktrace at the bottom of this issue. I have attached this file for reference. [spiral.seq.txt](https://github.com/user-attachments/files/23592987/spiral.seq.txt)

This issue seems to overlap somewhat with https://github.com/JuliaHealth/KomaMRI.jl/issues/323 and I can see that the code responsible for reading these parameters has been altered in https://github.com/JuliaHealth/KomaMRI.jl/pull/324, though it would be useful to have a fix implemented in the meanwhile before 3D reconstruction is implemented.

```
MethodError: no method matching AcquisitionData(::Dict{Symbol, Any}, ::Vector{MRIBase.Trajectory{Float32}}, ::Array{Matrix{ComplexF32}, 3}, ::Vector{Vector{Int64}}, ::Tuple{Float64, Float64}, ::Tuple{Float64, Float64, Float64})
The type `AcquisitionData` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  AcquisitionData(::Dict{Symbol, Any}, ::Array{MRIBase.Trajectory{T}, 1}, ::Array{Array{Complex{T}, 2}, 3}, ::Vector{Vector{Int64}}, ::NTuple{D, Int64}, ::Tuple{Float64, Float64, Float64}) where {T<:AbstractFloat, D}
   @ MRIBase ~/.julia/packages/MRIBase/UQbPt/src/Datatypes/AcqData.jl:21
  AcquisitionData(::ISMRMRDFile, ::Any)
   @ MRIFiles ~/.julia/packages/MRIFiles/9hXgx/src/ISMRMRD/ISMRMRD.jl:62
  AcquisitionData(::ISMRMRDFile)
[...]

PR #645, opened 2025-11-17
Added integer rounding for guessed recon dimensions

thePlasm:
This is a pull request to fix https://github.com/JuliaHealth/KomaMRI.jl/issues/644. With this change, I am able to successfully perform reconstruction for the spiral.seq file attached to that issue.


cncastillo:
Ok, there was a problem with the docs, that i fixed in #646. Testing again (don't mind the buildkite fail).

PR #658, opened 2025-12-03
Updating get_RF_types, get_RF_center, get_Mk

gsahonero:


KomaMRIBase/src/datatypes/sequence/RF.jl
```diff
@@ -189,8 +189,12 @@ RF delay.
...
-    t_center = sum(abs.(B1) .* t) ./ sum(abs.(B1))
+    if x.use != Undefined() &&  x.center != 0.0
+        t_center = x.center
+    else
+        t = times(x)
+        B1 = ampls(x)
+        t_center = sum(abs.(B1) .* t) ./ sum(abs.(B1))
+    end
     return t_center
 end
```
cncastillo: I think the idea is to dispatch based on the `RF{RFUse}`, and do exactly what we are doing now if the use is `Undefined()`, and return `r.center` otherwise. Here you are doing extra computations for the case that the RF center is defined.

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -425,25 +425,29 @@ Get RF centers and types (excitation or precession). Useful for k-space calculat
...
 
 # Returns
 - `rf_idx`: (`::Vector{Int64}`) indices of the RF centers
 - `rf_type`: (`::Vector{Int64}`, opts: [`0`, `1`]) RF type (`0`: excitation, `1`:
     precession)
 """
-function get_RF_types(seq, t)
-	α = get_flip_angles(seq)
+function get_RF_types(seq, t; ex_rf_th = 90.01)
+α = get_flip_angles(seq)
```
cncastillo: Don't add `ex_rf_th`, we need to force people to use the `RFUse`. The same logic as before, if `RFUse` is undefined, we need to do exactly what we were doing before.

Also, the computation of 

```julia
RF_ex = (α .<= ex_rf_th) .* RF_mask
RF_rf = (α .>  ex_rf_th) .* RF_mask
```

doesnt make sense if the RFUse is defined, and its being computed for all RFs.

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -425,25 +425,29 @@ Get RF centers and types (excitation or precession). Useful for k-space calculat
...
 	RF_mask = is_RF_on.(seq)
-	RF_ex = (α .<= 90.01) .* RF_mask
-	RF_rf = (α .>  90.01) .* RF_mask
+	RF_ex = (α .<= ex_rf_th) .* RF_mask
+	RF_rf = (α .>  ex_rf_th) .* RF_mask
 	rf_idx = Int[]
-	rf_type = Int[] #cambiar a tipo RFUse
+	rf_type = Vector{Any}()#Int[] #cambiar a tipo RFUse
```
cncastillo: Using containers of type any is a **very bad** practice, as it generates inefficient code. Please use `RFUse`,  if RFUse is undefined we do the same as before but replacing the ints 0 and 1 by `Excitation()` and `Refocusing()`).

KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -487,12 +491,18 @@ function get_Mk(seq::Sequence, k; Δt=1, skip_rf=zeros(Bool, sum(is_RF_on.(seq))
...
+						mk[p,i] .-= mkf		
+					else
+						mk[p,i] .+= mkf
+					end
+				elseif rf_type[rf-1] == 0 # Excitation
 					mk[p,i] .-= 0
 				elseif rf_type[rf-1] == 1 # Refocuse
 					mk[p,i] .-= mkf
-				#end
 				else
```
cncastillo: You have the logic repeated twice, this doesn't make sense, please only use an `rf_type::Vector{RFUse}` (like explained in previous comment).

cncastillo (changes requested):
Hi, thanks for the PR. I see some problems with the code, in general you need to take advantage of multiple dispatch. You can change RFs to be a parametric type `RF{RFUse}` for this and specialized functions based on the type of `RFUse`. We want to force the user to use Pulseq RFUse if they don't want us to guess the use of each RF with heuristics like alpha > 90. 

Look at MATLAB's Pulseq code to check how they deal with this for the calculation of k-space, so we can match behavior.

pvillacorta:
I addressed the requested changes in [8ce25d5](https://github.com/JuliaHealth/KomaMRI.jl/pull/658/commits/8ce25d56512d7281a33f6a15bfe9ad0412fc639c) and merged this PR into #614, so that we can continue from there.

PR #718, opened 2026-01-28
Write Pulseq 1.5

### LINKED ISSUE #152: Write Pulseq files

pvillacorta:
This PR fixes #152 and replaces #284 from @beorostica.

First of all, thanks to @beorostica for the original work and the time invested in that PR. Given the changes in master since then, I personally found it easier to reimplement the solution from scratch based on the current state of the master branch. The goal remains the same, with an updated implementation.

Round-trip tests (Koma `Sequence` → `write_seq` → .seq → `read_seq` → Koma `Sequence`) are still in progress. For now, a simple (`RF` + `Grad` + `ADC`) sequence is used, but it needs to be improved (more blocks, events, time-shaped waveforms, extensions, …).

Some additional functions and keyword arguments have been added compared to #614. We could take a deeper look at them in a meeting.


KomaMRIFiles/src/Sequence/Pulseq.jl
```diff
@@ -675,21 +678,24 @@ Block sequence definition. Used internally by [`read_seq`](@ref).
 - `i`: (`::Int64`) index of a block event
 - `pulseq_version`: (`::VersionNumber`) Pulseq version
 
+# Keywords
+- `simplify_shapes`: (`::Bool`) whether to simplify the shapes of the gradients and RFs
```
cncastillo: Is this `simplify_shapes` argument also available on Pulseq? Is that the same as waveform compression? If so I would try to use the same name.

KomaMRIFiles/test/runtests.jl
```diff
@@ -134,4 +134,16 @@ end
             end
         end
     end
-end
\ No newline at end of file
+    @testset "Round-trip Test" begin
+        include("utils.jl")
+        pth = (@__DIR__) * "/test_files/pulseq/"
+        # EPI sequence
+        filename = pth * "round_trip_test.seq"
```
cncastillo: I would I add more tests than this, only comparing one sequence is probably not robust enough.

I am particularly interested in covering:
- RF with phase offset (RF spoiled)
- RF with frequency modulation (phase waveform, it could be an adiabtic pulse)
- RF with frequency modulation + phase offset (RF spoiled with slice offset)
- Gradient uniformly sampled waveforms

I briefly mentioned that when having a phase offset, the RF center is important. And that comparing complex waveforms to check uniqueness is not trivial. I would like to see some tests on this area.

KomaMRIFiles/test/utils.jl
```diff
@@ -0,0 +1,35 @@
...
+    return qseq
+end
+
+function round_trip_seq()
+    seq = Sequence()
+    seq += RF(1.0, 100e-6)
+    seq += Grad(1.0, 100e-6, 10e-6)
+    seq += ADC(100, 10e-6, 10e-6)
+    return seq
+end
```
cncastillo: This seems too simple.

KomaMRIFiles/test/utils.jl
```diff
@@ -0,0 +1,35 @@
...
+
+quantize_time(t::Array, factor::Float64) = t
+quantize_time(t::Real,  factor::Float64) = round(Int, t / factor) * factor
```
cncastillo: Why `Real` and not `Number`. Don't overtype, `factor` doesn't need a type unless you are using it for dispatching.

cncastillo (changes requested):

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: It is not the same as waveform compression. This is Koma-specific and is intended to simplify `Grad` and `RF` events when possible. For instance, when an RF shape is defined in Pulseq as follows:

```
# Format of RF events:
# id ampl. mag_id phase_id time_shape_id center delay freqPPM phasePPM freq phase use
# ..   Hz      ..       ..            ..     us    us     ppm  rad/MHz   Hz   rad  ..
# Field 'use' is the initial of: 
#   excitation refocusing inversion saturation preparation other undefined
[RF]
1           25 1 2 3 5000 0 0 0 0 0 e

# Sequence Shapes
[SHAPES]

shape_id 1
num_samples 2
1
1

shape_id 2
num_samples 2
0
0

shape_id 3
num_samples 2
0
10000
```

The `read_seq` function would create the `RF` instance like this:
```julia 
rfAϕ = 25/γ .* [1, 1] .* exp.(1im*(2π*[0, 0]))
rf1 = RF(rfAϕ, [0, 10000] * rfRasterTime)
```
that is, an RF with a two-point waveform. 
However, for these cases where (a) the amplitude is constant along all time points and/or (b) all time points are equidistant, the RF instance can be simplified to:
```julia
rf2 = RF(25/γ, 10000 * rfRasterTime)
```
The `simplify_shapes` keyword controls whether or not this simplification is applied when possible. By default, it is set to `true`, but it is disabled in "Read Comparison" tests so all tests pass (`rf1` and `rf2` are not formally the same).

The alternative to this is to always simplify (remove `simplify_shapes` kwarg), and define the `≈` method in a way that `rf1` and `rf2` are treated as equal.

Let me know what you think.

cncastillo: Ok, simplifying makes sense (if this is explained somewhere). If its only simplifying this case, i would maybe call it differently.

KomaMRIFiles/test/utils.jl
pvillacorta: Solved

pvillacorta:
I have tried to cover most of the requested details. I think the only pending changes are:

- [x] Revise sequence comparison function:
    - [x] throw warning in case we are comparing via linear interpolation (when sequence samples don't have the same length).
    - [x] compare extensions
- [x] Documentation for specific functions or keywords (e.g., `simplify_shapes`). 
EDIT: I have decided to delete that keyword, since it is no longer needed. It's only purpose was to avoid simplifying waveforms when comparing sequences. Now, since the new implementation of `isapprox(::Sequence, ::Sequence)`, sequences can result in being equal even with waveforms with different number of points.
- [x] Ensure that samples of uniformly-shaped events are located at raster cell centers. This is needed in both read and write functions when no `time_shape` is provided (`time_sahape_id = 0`) (see [this](https://pulseq.github.io/specification.pdf#subsection.2.6)).
**EDIT: Due to differing sampling conventions (edge-based vs center-based) and integer microsecond delay quantization in Pulseq, compact uniformly-sampled shapes (`time_shape_id = 0`) are not used to avoid introducing temporal inaccuracies. Therefore, it is possible to read sequences with uniformly-shaped events, but they will be converted to "time-shaped" events once they are written again. This approach may increase file size, but ensures that we are not making any error derived from quantization.**
- [x] More tests (extensions, edge cases...)
- [x] Bump package versions (KomaBase v0.9.10? KomaFiles v0.9.12?)

KomaMRIFiles/src/Sequence/Pulseq.jl
pvillacorta: Deleted this keyword. No longer needed. Simplification is always applied.

cncastillo:
Thanks! I tested it on this sequence: 
[fatsat_r2_5mm_mtx100x100_0_05s.seq.txt](https://github.com/user-attachments/files/26910307/fatsat_r2_5mm_mtx100x100_0_05s.seq.txt)

Read takes 6ms, write takes 70ms (~10x). 

I also did a "roundtrip" test: `read` -> `write` -> `read` and it gave me an error. I think the problem was the trigger extension. The writer saves the trigger like:
```
1 2.0 1.0 0.0 0.00000053
```
but the reader expect `%i %i %f %f`. For reference, in the original file we had:
```
1 2 1 0 530000 
```
so there is an error in the type and unit conversions as well.

Also, the saved DEFINITIONS are weird (we may want to ignore some `# <---`):
```
DEFINITIONS]
signature (type = "md5", hash = "e8f4f8ab105b4b52470b2c0a3b6e1e09") # <-------------
AdcRasterTime 2.0e-6
GradientRasterTime 1.0e-5
FOV [0.25, 0.25, 0.01]
Name stanford_logo
Nz 1
Nx 100
Ny 100
PulseqVersion 1.5.0 # <-------------
BlockDurationRaster 1.0e-5
FileName bSSFP_fatsatopt_r2_5mm_mtx100x100_0_05s.seq
RequiredExtensions String[] # <-------------
RadiofrequencyRasterTime 1.0e-6
```

cncastillo:
Also, right now there are some pretty big differences on the output format. Like hard coding two spaces `"  "` between columns in the block table and others things like that, instead of using `printf` like MATLAB's. Not really a big problem, can be fixed later.

I would also suggest to split Pulseq.jl to PulseqRead.jl and PulseqWrite.jl.

pvillacorta:
I made a set of improvements related to your comments:
- Output formatting; this change has been guided by Cursor, so it can certainly be simplified. For the moment, it works well. (aac0ec6dfaeaeefe1416a4fb213bc22a9de1b9d8)
- Fix Trigger output format. (aac0ec6dfaeaeefe1416a4fb213bc22a9de1b9d8)
- Optimize writing by using an array for storing blocks and hash tables for shapes and events (c36121d9d0af3957a65c4f311761281c998dd680).
- Clean definition export. (0b10a5b9a796190a7e5966ab6baa65b339b04961)
- Add `_slew_quotient_ΔG_over_Δt` function for slew rate calculations and update `get_slew_rate` and `get_eddy_currents` to use it. Now it does not produce extremely high values (I don't know why you did not get an error when exporting [fatsat_r2_5mm_mtx100x100_0_05s](https://github.com/user-attachments/files/26910307/fatsat_r2_5mm_mtx100x100_0_05s.seq.txt)). (760b74e3a6ff855cd99f2e62720584f21b2bfd13)
- File and folder redistribution as requested. (034f48d18a18f21b35b3a3e7e7d89f4f6c646594)

cncastillo:
I will close this one in favor of #754 which is already complete! thanks for your efforts!

PR #759, opened 2026-04-29
Add start and end points to the sequence time points calculated by `KomaMRIBase.get_variable_times`

### LINKED ISSUE #758: `get_variable_times` throws BoundsError for pure Delay sequences
### Description

  `KomaMRIBase.get_variable_times` throws a `BoundsError` when called on a
  sequence that contains only delay blocks, for example `Sequence() +
  Delay(1.0)`.

  This appears to happen because pure delay blocks do not add any RF, gradient,
  ADC, or motion timing points. The internal `t` vector remains empty, and the
  function later indexes `t[1]` / `t[end]` while adding its endpoint padding.

 ### Minimal Reproducer

  ```julia
  using KomaMRIBase

  seq = Sequence() + Delay(1.0)
  t, Δt = KomaMRIBase.get_variable_times(seq)
```

### Current Behavior

`ERROR: BoundsError: attempt to access 0-element Vector{Float64} at index [1]`

The failing code path is reached after no timing points are added for the delay-only sequence.

### Expected Behavior

`get_variable_times` should return a valid time grid for a pure delay sequence, e.g. a coarse interval spanning the sequence duration, rather than throwing an error.

For example, a reasonable result would include the delay start and end times, with the existing small endpoint padding convention applied.

### Notes

This affects pure-delay sequences without RF, gradients, ADC, or motion key times. A minimal fix is to add `[0.0, dur(seq)]` as fallback timing points when the collected time vector is empty before the existing `sort!(unique!(t))` and endpoint-padding logic.

Given a pure delay sequence is of little interest in practice, the severity of this bug is very low and the fallback code path would likely almost never be reached. This issue surfaced during the coding of unit tests for a new simulation method. Please let me know if tiny bugs like these are worth reporting or if this just adds noise.

The text of this bug report was drafted by Codex (GPT-5.5) and amended by JanWP.

### Environment

```Markdown
OS x86_64-linux-gnu
Julia 1.12.6
KomaMRIPlots 0.9.6
KomaMRIFiles 0.9.11
KomaMRI 0.9.2
KomaMRICore 0.9.6
KomaMRIBase 0.9.9
```

JanWP:
This closes #758.


cncastillo (approved):
This looks good to me. I will just merge it.

PR #765, opened 2026-05-02
Reject negative ADC labels and default raw recon to custom

### LINKED ISSUE #637: Catch negative labels in sequences
### Feature Request

Hi Devs,

I am working with some 4DFlow sequences, and am using labels for ky,kz and cardiac phase/encoding. 
I accidentally included negative labels for ky,kz while debugging, and this results in a crash after simulation as the ISMRMRD format uses `uint16`. 

In this case it was not too much of a problem, as the simulation was only a few minutes long, but this could be annoying for users who want to perform very long simulations, as it happens after the simulation.

It would be great if `simulate()` checked the sequence for negative labels at the start, and threw an error prior to any significant computation being done. Additionally, maybe it would be useful if `read_seq()` also gave a warning when importing a sequence with negative labels.

cncastillo:
## Problem
Negative ADC labels fail late during raw-data conversion because ISMRMRD encoding counters are unsigned. For long simulations this means users only see the failure after simulation work has already completed.

The raw-data header also used `other` as the default trajectory label, while KomaMRI-generated raw data should use `custom` for reconstruction.

## Solution
- Checks cumulative ADC labels with `get_labels(seq)` at the start of `simulate` and stops on the first negative value.
- Changes the raw-data trajectory default from `other` to `custom`.
- Updates the label tutorial to start from the default zero counter instead of using a negative `LabelSet` hack.

## Tests
- `Pkg.test("KomaMRICore")`

## Development
Closes #637
Context: discussion #786

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -156,6 +156,9 @@ jobs:
     if: '!cancelled()'
     name: Documentation
     runs-on: ubuntu-latest
+    concurrency:
+      group: docs-gh-pages
+      cancel-in-progress: false
     steps:
       - uses: actions/checkout@v6
       - uses: julia-actions/setup-julia@v2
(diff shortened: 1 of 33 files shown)


cncastillo:
CI failed in the docs build because `examples/3.tutorials/lit-07-label.jl` used the old `LabelSet(-1, "LIN")` offset trick before simulation. This branch now assigns explicit zero-based nonnegative `LIN` values on ADC blocks, so the validation remains strict and the tutorial still feeds MRIReco zero-based lines.

Verified locally with `julia --project=docs docs/make.jl`.

cncastillo:
Follow-up: the tutorial now uses the intended `LabelInc` form instead of explicit `LabelSet` values. Label counters already start at 0; the increment is attached to the block after each ADC, so ADC profiles read `0, 1, 2, ...` without ever creating a negative label state.

Verified with `julia --project=docs docs/make.jl`.

PR #787, opened 2026-05-21
fix: optimize Rotate displacement! matrix operations (#604)

### LINKED ISSUE #604: Optimize Rotation motion

Mxyank24:
Fixes #604

Refactored Rotate displacement functions to use loop fusion (@.) and deg2rad. This eliminates intermediate array allocations (e.g., x .- cx) in the hot loop, resolving the 10x slowdown.

### DIFF
--- a/KomaMRIBase/Project.toml
+++ b/KomaMRIBase/Project.toml
@@ -1,7 +1,7 @@
 name = "KomaMRIBase"
 uuid = "d0bc0b20-b151-4d03-b2a4-6ca51751cb9c"
 
-version = "0.11.1"
+version = "0.11.2"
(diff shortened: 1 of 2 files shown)


cncastillo:
Thank you so much! :) waiting for tests to pass and we will register a new version of KomaMRIBase.

Mxyank24:
> Thank you so much! :) waiting for tests to pass and we will register a new version of KomaMRIBase.

My pleasure sir.

Issue #788, opened 2026-05-23
Residual per-shot signal drift past ~270 s (same FP collapse as #780, two more sites)


ArthurAllilaire:
### What happened?

#780 fixed the rising-edge FP-epsilon collapse, but the same hazard remains at the time-rebasing sites. `_separate_closing_knot!` separates the closing knot by a fixed `1e-14` in *block-relative* coordinates; `get_samples` and `get_variable_times` then rebase to absolute time with `T0 .+ t`. Once `eps(T0 + t[end]) > 1e-14` (T0 ≳ 270 s for ms-scale events), the gap collapses, the closing knot equals the block-end knot, and the knots are removed by `unique!` leading to the same errors as last time.

## Fix

Leave the block-relative `_separate_closing_knot!` untouched, and re-separate the closing knot on the rebased absolute vectors in `get_samples` / `get_variable_times` with an idempotent adaptive gap `max(MIN_RISE_TIME, eps(t[end]))` that widens past `1e-14` once `t[end]` is large.

## Reproducer

24 identical IR-90° shots, TR = 15 s, no gradients — every `|signal|` should be identical, but shot 18 (sim_time = 270 s) jumps +52 %:

```
shot 1..17 (≤255 s): |signal| = 0.476267   stable
shot 18    (270 s):  |signal| = 0.722479   +52 %
shot 19..24(≥285 s): |signal| = 0.677159   new wrong plateau
```

```julia
using KomaMRI, Suppressor, Printf

const B1_AMP  = 20e-6     # 20 µT → 180° in 1 ms
const T_RF    = 1e-3
const T_ADC   = 1e-3
const TI      = 3.0
const TR      = 15.0
const N_SHOTS = 24

obj = Phantom(name="spin", x=[0.0], y=[0.0], z=[0.0],
              T1=[1.0], T2=[1.0], T2s=[1.0], ρ=[1.0], Δw=[0.0])

function rf_block(α, T)
    amp = B1_AMP * (α / π)
    Sequence(reshape([Grad(0.0, T), Grad(0.0, T), Grad(0.0, T)], 3, 1),
             reshape([RF(amp, T)], 1, 1), [ADC(0, 0.0)])
end

function adc_block(T)
    Sequence(reshape([Grad(0.0, T), Grad(0.0, T), Grad(0.0, T)], 3, 1),
             reshape([RF(0.0, T)], 1, 1), [ADC(1, T)])
end

function build_seq(N; TI, TR)
[...]

PR #789, opened 2026-05-24
Fix FP knot collapse at large absolute time

ArthurAllilaire:
## Summary

Fix for #788. `_separate_closing_knot!` separates the closing knot by a fixed `1e-14` in block-relative coordinates, but `get_samples` and `get_variable_times` rebase to absolute time (`T0 .+ t`). Once `eps(T0 + t[end])` exceeds `1e-14` (T0 ≳ 270 s for ms-scale events), the gap collapses.

## Fix

I first tried making `_separate_closing_knot!` itself adaptive, but that changes the block-relative knots from `times()` and broke the gradient-addition test. So I split it: `_separate_closing_knot!` stays exactly as it was, and a new `_reseparate_closing_knot!` does the adaptive re-separation on the rebased absolute vectors at every place I could find where a rebase happens (`get_samples`, `get_variable_times`):

```julia
_reseparate_closing_knot!(t) =
    t[end-1] = min(t[end-1], t[end] - max(MIN_RISE_TIME, eps(t[end])))
```

It is **idempotent** (important, since the rebased vectors have already been through `_separate_closing_knot!` once) and only pulls a collapsed closing knot down to `t[end] - gap`, leaving an already-separated knot alone. The adaptive `gap` is `MIN_RISE_TIME` at small magnitudes (near-no-op at small `T0`) and widens to ≥ 1 ULP at large absolute time, which is what survives the rebasing.

For gradients the rebasing site also re-applies `_strictly_increasing_knots!` to lift leading/internal duplicates up. RF only needs the closing pass; its leading duplicate is handled by `deduplicate_knots!` in `get_rfs`.

## Verification

- `koma_bug_residual.jl`: 24 identical IR shots, TR=15 s → all `|signal|` identical (was +52 % at shot 18).
- `Pkg.test("KomaMRIBase")` green (544/544), including the #780 testset, the gradient-addition linearity test, and the k-space rotation tests.

P.S. Have tried to make this less verbose!


ArthurAllilaire:
@cncastillo

cncastillo:
Sorry about that! I have been very busy lately and I missed this one, i will merge as soon as I merge #799.

cncastillo (approved):

PR #790, opened 2026-05-24
Add button in KomaUI to reload sequence

mcgrathcm:
Currently in KomaUI, if the underlying .seq file is changed, it is only possible to reload the sequence by first loading a different sequence, then loading the one you want.
This just adds a button to the left of the sequence picker to allow the user to reload the sequence.
Very useful when coupled with the sequence generation functionality in Koma.


cncastillo:
That is cool! do you have a preview?

mcgrathcm:
Currently on the left side. Could also move it to the left of the sequence name. 

image

cncastillo:
I also tried it to the right (and removed the very old Koma seq)
image

What do you think?

What is the workflow that this could help? It would be really nice if the thing just detected changes in the file and reloaded it when changed, but this is a good intermediate step.

mcgrathcm:
That looks better to me, makes more sense on the right.
In terms of workflows, I think it mostly helps for sequence debugging and tuning, since I can adjust and run a sequence building script and then check it very quickly.

PR #794, opened 2026-06-01
(WIP) Phantom time dependent properties

pvillacorta:
This PR started as I needed to simulate variability in the `ρ` field of the Phantom. I then extended it to the rest of vector fields (T1, T2, ...). For now, it is functional, but it is still a work in progress. I just leave it here so anyone interested could use it or improve it so far. 

Additionally, some thoughts came to my mind related to change the way motion is defined. Maybe we could include motion directly as trajectories within the `x`, `y` and `z` fields, and avoid using a dedicated `motion` field. We might think a bit more in depth regarding this.


KomaMRIBase/src/datatypes/Phantom.jl
```diff
@@ -95,20 +104,28 @@ end
...
+        fields...,
+        motion=vcat_motion(obj1.motion, obj2.motion, length(obj1), length(obj2)),
+    )
 end
 
 """Scalar multiplication of a phantom"""
 *(α::Real, obj::Phantom) = begin
     obj1 = deepcopy(obj)
-    obj1.ρ .*= α
+    if obj1.ρ isa AbstractVector
```
cncastillo, line 124: i would try not to do this

KomaMRIBase/src/timing/Interpolation.jl
```diff
@@ -54,6 +54,10 @@ function resample(itp::Interpolator2D, t)
     Ns = size(itp.coefs, 1)
     id = _similar(t, Ns)
     copyto!(id, collect(range(oneunit(eltype(t)), eltype(t)(Ns), Ns)))
+    if t isa AbstractVector
```
cncastillo, line 57: same here

cncastillo (commented):
Thanks! Just a few comments. I am a little worried in performance, we would need to check that the "normal" case is not affected. Probably the Phantom would need to use parametric types so the compiler can see if the property has motion or not.

PR #799, opened 2026-06-12
Update Buildkite pipeline for the new JuliaGPU cluster

maleadt:
The JuliaGPU Buildkite agents have moved to a dedicated cluster, where the single `juliagpu` queue has been split into per-backend queues. Steps now select agents using `queue: "cuda"`, `queue: "rocm"` or `queue: "oneapi"` instead of `queue: "juliagpu"` combined with a `cuda`/`rocm`/`intel` tag. Steps that targeted `queue: "juliagpu"` without a backend tag now use the cluster's default queue.

The CUDA benchmark step used the old `benchmark` queue, which is gone; benchmark agents now live in the backend queues with a `benchmark` tag, so that step now targets `queue: "cuda"` with `benchmark: "*"` (keeping the `gpu: "rtx2070"` selector).

Queues are cluster-scoped, so steps targeting `queue: "juliaecosystem"` (the CPU aarch64 test/benchmark steps and the macOS Metal test/benchmark steps) can no longer run as part of the GPU-cluster pipeline. They have been moved verbatim into `.buildkite/pipeline-julia.yml`, to be wired up to a separate pipeline in the ecosystem cluster once it is back online. Until then, the "Combine benchmarks" aggregation step will temporarily not include CPU and Metal results, since those artifacts are no longer produced within the same build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.buildkite/pipeline-julia.yml
+++ b/.buildkite/pipeline-julia.yml
@@ -1,0 +1,136 @@
+steps:
+  - group: ":julia: ($TEST_GROUP) Tests"
+    steps:
+      - label: "CPU 4T - Julia {{matrix.version}} - Test"
+        matrix:
+          setup:
+            version:
+              - "1.10"
+              - "1"
+        plugins:
+          - JuliaCI/julia#v1:
+              version: "{{matrix.version}}"
+        command: |
+          julia --project=KomaMRICore -e 'println("--- :julia: Instantiating project")
+              using Pkg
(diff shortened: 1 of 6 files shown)


cncastillo:
Thank you very much! The CPU benchmarks not being available messes with our aggregation step, so I had to update that one to be more robust. Running tests and merging ASAP.

PR #816, opened 2026-07-08
Hoist constants in GPU Bloch kernels

KookiesNKareem:
## Summary

This PR hoists repeated scalar work out of the benchmarked GPU Bloch and Bloch-Magnus kernel loops.

The PR was intentionally opened from an initial trigger commit first so the `run-gpu-ci` label could be applied before the performance changes were committed and pushed.

## Changes

- Precompute `-2πγ`, `1/γ`, and relaxation reciprocals outside per-step GPU loops.
- Precompute the GL4 `B_to_ω^2 * sqrt(3)` scale once per kernel invocation.
- Add compatible `rotation_vector` overloads that accept precomputed constants while keeping the previous call signatures working.
- Hoist the precession phase scale in the GPU precession kernel.

## Validation

- `git diff --check`
- `Pkg.test("KomaMRICore"; test_args=["CPU"])`: 431 passed
- KernelAbstractions CPU-backend smoke compile/run for the patched GPU-style excitation kernels
- CUDA microbenchmark on Ultron over the six `run-gpu-ci` paths: 1.0446x aggregate median speedup
- Local Metal microbenchmark was noisy/inconclusive, so CUDA is the cleaner performance signal here

### DIFF
--- a/KomaMRICore/src/simulation/SimMethods/Bloch/gpu/ExcitationKernel.jl
+++ b/KomaMRICore/src/simulation/SimMethods/Bloch/gpu/ExcitationKernel.jl
@@ -47,6 +47,8 @@ end
     i_g = @index(Group, Linear)
     i = (i_g - 1u32) * UInt32(N) + i_l
 
+    B_to_ω = T(-2π * γ)
+    inv_γ = inv(T(γ))
     sig_group_r = @localmem T HAS_ADC ? (USE_WARP_REDUCTION ? 32 : N) : 1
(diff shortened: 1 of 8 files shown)


KookiesNKareem:
Closing and recreating from a same-repository branch so GPU benchmark CI is not skipped by fork guards.

christiangnrd:
I cancelled the buildkite runs since this was closed/deleted and reopened. If I shouldn't have, let me know and I'll retry the jobs that got interrupted

KookiesNKareem:
Thank you for doing that, no problems

PR #817, opened 2026-07-08
Hoist constants in GPU Bloch kernels

KookiesNKareem:
## Summary

This PR hoists repeated scalar work out of the benchmarked GPU Bloch and Bloch-Magnus kernel loops.

The PR was opened from a same-repository branch with an initial trigger commit first so the `run-gpu-ci` label could be applied before the performance changes were committed and pushed.

## Changes

- Precompute `-2πγ`, `1/γ`, and relaxation reciprocals outside per-step GPU loops.
- Precompute the GL4 `B_to_ω^2 * sqrt(3)` scale once per kernel invocation.
- Add compatible `rotation_vector` overloads that accept precomputed constants while keeping the previous call signatures working.
- Hoist the precession phase scale in the GPU precession kernel.

## Validation

- `git diff --check`
- `Pkg.test("KomaMRICore"; test_args=["CPU"])`: 431 passed
- KernelAbstractions CPU-backend smoke compile/run for the patched GPU-style excitation kernels
- CUDA microbenchmark on Ultron over the six `run-gpu-ci` paths: 1.0446x aggregate median speedup
- Local Metal microbenchmark was noisy/inconclusive, so CUDA is the cleaner performance signal here


KomaMRICore/src/simulation/SimMethods/Bloch/gpu/ExcitationKernel.jl
```diff
@@ -58,6 +59,8 @@ end
     ΔBz = zero(T)
     T1 = T(1)
     T2 = T(1)
+    neg_inv_T1 = T(-1)
+    neg_inv_T2 = T(-1)
```
cncastillo: is that just `-1`? i don't think it makes so much sense to pre-compute that one

KomaMRICore/src/simulation/SimMethods/Bloch/gpu/ExcitationKernel.jl
```diff
@@ -72,6 +75,8 @@ end
         ρ = p_ρ[i]
         T1 = p_T1[i]
         T2 = p_T2[i]
+        neg_inv_T1 = -inv(T1)
+        neg_inv_T2 = -inv(T2)
```
cncastillo:
```julia
neg_inv_T1 = -inv(p_T1[i])
neg_inv_T2 = -inv(p_T2[i])
```
Should use less registers.

KomaMRICore/src/simulation/SimMethods/Bloch/gpu/PrecessionKernel.jl
```diff
@@ -50,7 +51,7 @@
             Δt = s_Δt[s_idx-1]
             t += Δt
             Bz_next = x * s_Gx[s_idx] + y * s_Gy[s_idx] + z * s_Gz[s_idx] + ΔBz
-            ϕ += (Bz_prev + Bz_next) * T(-π * γ) * Δt
+            ϕ += (Bz_prev + Bz_next) * precession_scale * Δt
```
cncastillo: `B_to_ω_half`

cncastillo (changes requested):

PR #919, opened 2026-07-23
Skip GPU excitation relaxation for infinite T1 and T2

KookiesNKareem:
## Summary

- detect phantoms whose T1 and T2 values are all infinite before GPU transfer
- carry that result through GPU preallocation as a `Val` type parameter
- specialize the excitation kernel so the infinite-relaxation path omits T1/T2/rho loads and both exponential evaluations

## Why

For infinite T1 and T2, E1 and E2 are exactly one. The GPU excitation kernel was still loading relaxation data and evaluating two exponentials for every spin and time step. Encoding the condition in the kernel type gives the GPU compiler separate finite- and infinite-relaxation kernels while keeping a single kernel implementation.

Finite-relaxation behavior is unchanged.

## Validation

- `Pkg.test("KomaMRICore")`: 432/432 passed
- Metal A/B, 10,000 spins x 13 steps, synchronized, 100 samples x 100 launches:
  - regular: 70.78 us/launch
  - infinite-relaxation fast path: 57.05 us/launch
  - speedup: 1.24x
  - maximum Mxy/Mz error: 0
- `git diff --check`

The `run-gpu-ci` label is attached for CUDA coverage.

### DIFF
--- a/KomaMRICore/src/simulation/SimMethods/Bloch/gpu/BlochGPU.jl
+++ b/KomaMRICore/src/simulation/SimMethods/Bloch/gpu/BlochGPU.jl
@@ -3,7 +3,7 @@ include("PrecessionKernel.jl")
 include("ExcitationKernel.jl")
 
 """Stores preallocated arrays for use in Bloch GPU run_spin_precession! and run_spin_excitation! functions."""
-struct BlochGPUPrealloc{T} <: PreallocResult{T}
+struct BlochGPUPrealloc{T,SKIP_RELAXATION} <: PreallocResult{T}
     sig_output::AbstractMatrix{Complex{T}}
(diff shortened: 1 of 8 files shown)


cncastillo:
Actually we might be able to make it even faster for the excitation kernel. For the non-relaxation case, the excitation is equivalent to Spinor reduction.

$$
(\alpha_1, \beta_1) \times \dots \times (\alpha_N, \beta_N) = (\alpha_{eq}, \beta_{eq})
$$

So we don't need repeated rotations of $\vec{M}$, we just do a parallel reduction on the spinors and then apply $(\alpha_{eq}, \beta_{eq})$ once, or at least once per ADC sample (generally there are no ADCs during excitation).

PR #920, opened 2026-07-29
Include citation to motion paper in README.md

pvillacorta:
Include citation to: https://doi.org/10.1002/mrm.70145,
so that people using motion functionalities can cite both the original paper and this one.


cncastillo:
Perfect!

PR #924, opened 2026-07-31
Enable Reactant AD through sequence discretization

KookiesNKareem:
## Summary

- make single-block sequence discretization traceable through Reactant, including RF waveform interpolation and ADC-at-start padding
- add a narrow `KomaMRIBase` Reactant extension for traced RF arrays while preserving the existing native discretization path
- differentiate an acquired-signal loss through `discretize`, `get_sim_ranges`, and the `BlochSimple` time iterator

## Validation

- focused CPU Reactant + Enzyme ADC test: `16/16`
- `KomaMRIBase`: `767/767`
- full unfiltered `KomaMRICore` CPU suite: `448/448`
- `KomaMRIFiles`: `24448/24448`
- Titan RTX CUDA: existing parallel BlochSimple Reactant + Enzyme test `4/4`; new discretize + ADC + iterator test `2/2`
- CUDA gradient matched a fifth-order central finite-difference baseline at `rtol=1e-8`, `atol=1e-10`
- `git diff --check`

Stacked on #922.

PR #925, opened 2026-07-31
Enable Reactant AD through simulate

KookiesNKareem:
## Summary

- keep `simulate` signal and state storage traceable instead of unconditionally materializing CPU arrays
- differentiate an RF-scale loss through the public `simulate` pipeline, including sequence discretization, ADC sampling, and `BlochSimple`
- cover both Reactant CPU and CUDA execution with Enzyme gradients checked against finite differences

## Stack

This is PR 4 of 4, stacked directly on #924. Review only commit `a57b59092` for this layer.

## Validation

- `TEST_GROUP=reactant Pkg.test("KomaMRICore")`: 16/16 passed
- `Pkg.test("KomaMRICore")`: 448/448 passed
- Titan RTX via Reactant/XLA: kernel path 4/4 passed
- Titan RTX via Reactant/XLA: public `simulate` + ADC path 2/2 passed

### DIFF
--- a/KomaMRICore/src/simulation/SimulatorCore.jl
+++ b/KomaMRICore/src/simulation/SimulatorCore.jl
@@ -384,7 +384,8 @@ function simulate(
     if sim_params["gpu"]
         sim_params["Nthreads"] = 1
     end
-    sig = zeros(ComplexF64, Ndims..., sim_params["Nthreads"])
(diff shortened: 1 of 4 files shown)


KookiesNKareem:
Superseded by #926, which was created from the upstream-hosted branch through GitHub's native stacked pull request flow.

PR #922, opened 2026-07-31
Enable Reactant AD through parallel spin kernels

KookiesNKareem:
## Summary

- bypass `ThreadsX` and view construction when the parallel excitation/precession wrappers run with one worker, which is the GPU and Reactant path
- construct the spinor imaginary constant in the input floating type and skip empty precession acquisition work
- add a focused Reactant + Enzyme reverse-mode regression test against a fifth-order central finite-difference baseline

## Validation

- Reactant loss: `0.6920155690793802` (identical to native)
- Reactant + Enzyme gradient: `[0.019503560659149446, 0.0196742608887282, 0.0]`
- finite-difference gradient: `[0.01950356065917826, 0.019674260888777644, -1.1922786455233403e-15]`
- CPU core tests: non-motion `384/384`; motion `50/50`
- Metal core tests on Apple M4: non-motion `384/384`; motion `50/50`
- final focused Spinor/BlochSimple/AD Metal checks: `69/69`

### Metal benchmark

`BlochSimple`, official benchmark fixtures, 30-second BenchmarkTools trials, explicit `Metal.synchronize()`, two warm runs per source state:

| Fixture | #814 | This PR | Average delta |
| --- | ---: | ---: | ---: |
| Slice Selection 3D | 0.858–0.923 s | 0.862–0.891 s | -1.5% |
| MRI Lab | 0.715–0.762 s | 0.642–0.689 s | -10.0% |

Stacked on #814.

PR #814, opened 2026-07-07
Remove unnecessary {T}

KookiesNKareem:
## Summary

Remove the precision-defensive `{T<:Real}` parameter from public and internal
structs across `KomaMRIBase` and `KomaMRICore`. Method bodies that still need
the eltype now derive it inline (`T = eltype(p.ρ)`) instead of threading it
through every signature. This is a step towards native KomaMRI AD compatibility.

### Structs that lost `T`
- **Public API** (KomaMRIBase): `Phantom`, `Motion`, `MotionList`, `TimeCurve`,
  the `AbstractAction`/`SimpleAction`/`ArbitraryAction` hierarchy, and all
  concrete actions (`Translate`, `Rotate`, `HeartBeat`, `Path`, `FlowPath`).
- **Internal** (KomaMRICore): `DiscreteSequence`, `Mag`, `Spinor`,
  `SpinStateRepresentation`, `PreallocResult` and all concrete prealloc
  structs (`BlochCPUPrealloc`, `BlochMagnusCPUPrealloc`, `BlochGPUPrealloc`,
  `DefaultPrealloc`).

### Method signatures
Dropped `where {T<:Real}` (and associated `Phantom{T}` / `Complex{T}` /
`AbstractVector{T}` plumbing) from orchestrators in `SimulatorCore`, the
sim-method `run_spin_*!` paths, `trapz`/`cumtrapz`, scalar `Spinor`
constructors, `get_spin_coordinates`, `QuaternionRot`, and
`initialize_spins_state`. Precision is enforced upstream by the
`seqd |> f32`/`f64` pipeline.

### What stays
- GPU kernels (`precession_kernel!`, `excitation_kernel!`) still carry `T`
  for `@localmem T` and per-element `T(γ)` / `Complex{T}(im)` literals —
  replacement is a follow-up (precision-aware constants).
- Same-subtype constraints on `Extension`/`Grad` equality and `vcat`, plus
  generic-type plumbing in `Sequence._copy_events`/`_event_tuple` — not
  precision-related, load-bearing.

### Validation
- `KomaMRIBase`: 537/537
- `KomaMRICore` `:core`: 147/147
- AD suite: 5/5

PR #926, opened 2026-07-31
Enable Reactant AD through simulate

KookiesNKareem:
## PR Stack Summary

This is the final PR in a stack of 4 PRs to enable Reactant tracing + Enzyme AD through KomaMRI's high-level `simulate` function. Making `simulate` differentiable is the culmination of a lot of work that went into making Koma's simulations compatible with Enzyme, enabling downstream work like RF pulse optimization and physics-based MRI reconstruction. Coupled with Reactant XPU compilation, this lets Koma power highly efficient, differentiable-physics-based optimizations.

[PR 1](https://github.com/JuliaHealth/KomaMRI.jl/pull/814)'s primary objective was to remove overly strict typing in Koma that breaks Reactant's ability to trace through various key functions. GPU precision types are preserved upstream, and the PR's changes were checked to ensure that there were no serious regressions in type stability or performance.

[PR 2](https://github.com/JuliaHealth/KomaMRI.jl/pull/922) enabled Reactant + Enzyme AD through the base physics kernels: `run_spin_precession!` and `run_spin_excitation!`. This primarily included bypassing `ThreadsX` and view construction when they were launched with only 1 thread.

[PR 3](https://github.com/JuliaHealth/KomaMRI.jl/pull/924) extended Reactant traceability from the physics kernels through sequence discretization and simulation-range construction. It preserves traced array types through RF padding, waveform interpolation, and `DiscreteSequence` assembly. It also removes value-dependent branching from RF-center calculation and replaces findall-based range splitting. 

This PR completes the stack by enabling Reactant + Enzyme reverse-mode AD through KomaMRI’s public simulate function. It preserves Reactant-traced arrays throughout the high-level simulation pipeline, carries gradients through RF-node interpolation and both the `BlochSimple` and `Bloch` simulation methods, and adds CPU and CUDA checks against finite differences. I also added a tutorial to demonstrate slice-selective RF pulse design using AD.


KomaMRIBase/src/datatypes/Sequence.jl
```diff
@@ -283,6 +283,19 @@ Base.getindex(x::Sequence, i::AbstractVector{Bool}) = view(x, i)
...
+Return a sequence with RF event `index` replaced by an otherwise identical RF
+event whose amplitude is `A`.
+"""
+function set_rf_amplitude(seq::Sequence, A, index::Integer=1)
+    pulses = Matrix{RF}(seq.RF)
+    rf = pulses[index]
+    pulses[index] = RF(A, rf.T, rf.Δf, rf.delay, rf.center, rf.ϕ, rf.use, Val(:preserve))
+    return Sequence(seq.GR, pulses, seq.ADC, seq.DUR, seq.EXT, seq.DEF)
+end
+
```
cncastillo: i dont undestand why we need to do this, and not just use `seq.RF[idx].A = something`, I am strongly against defining set/get again.

examples/3.tutorials/lit-12-ReverseModeSignalMatching.jl
```diff
@@ -0,0 +1,100 @@
...
+    T2=fill(80e-3, 3),
+    Δw=2π .* [-120.0, 0.0, 120.0],
+)
+sim_params = Dict{String,Any}(
+    "sim_method" => Bloch(),
+    "gpu" => false,
+    "Nthreads" => 1,
+    "return_type" => "mat",
+    "precision" => "f64",
+    "sampling_rule" => MaxStepSizeRule(50e-6, 25e-6),
```
cncastillo: you dont need to define most of this. also use `Δt` and `Δt_rf`, `MaxStepSizeRule` is internal.

examples/3.tutorials/lit-11-ReverseModeRFOptimization.jl
```diff
@@ -0,0 +1,78 @@
...
+    obj=Phantom(; x=zeros(9), z=collect(range(-4e-3, 4e-3; length=9))),
+    sys=Scanner(),
+    sim_params=Dict{String,Any}(
+        "sim_method" => Bloch(),
+        "gpu" => false,
+        "Nthreads" => 1,
+        "return_type" => "state",
+        "precision" => "f64",
+        "sampling_rule" => MaxStepSizeRule(50e-6, 25e-6),
+    ),
```
cncastillo, line 87: you dont need to define most of this. also use `Δt` and `Δt_rf`, `MaxStepSizeRule` is internal.

examples/3.tutorials/lit-11-ReverseModeRFOptimization.jl
```diff
@@ -0,0 +1,78 @@
+# # Reverse-mode RF pulse optimization
+#
+# Optimize three RF control points through interpolation and a Bloch simulation
```
cncastillo: Tutorials are used to teach something. What is the MRI teaching goal with this one (?). The tutorial needs to be completely restructured.

Let's not confuse the method with the goal. Check https://diataxis.fr/tutorials/. This one is a good example https://juliahealth.org/KomaMRI.jl/dev/tutorial/gen-07-DiffusionSignalAttenuation, the method was the new `Path` motion, but its used to explain diffusion MRI signal attenuation.

Names of variables need to match the maths!!, A(x), b, ..., i don't see formulas.

In the code, use names like `forward` and `loss` not `signal_loss` and `signal_forward` or stuff like that. That doesn't make sense. Use the simplest name based on the context.

AI is very bad at creating this type of documentation page, we need more human intervention.

examples/3.tutorials/lit-12-ReverseModeSignalMatching.jl
```diff
@@ -0,0 +1,100 @@
+# # Reverse-mode signal matching
```
cncastillo: Tutorials are used to teach something. What is the MRI teaching goal with this one (?). The tutorial needs to be completely restructured.

Let's not confuse the method with the goal. Check https://diataxis.fr/tutorials/. This one is a good example https://juliahealth.org/KomaMRI.jl/dev/tutorial/gen-07-DiffusionSignalAttenuation, the method was the new `Path` motion, but its used to explain diffusion MRI signal attenuation.

Names of variables need to match the maths!!, A(x), b, ..., i don't see formulas.

In the code, use names like `forward` and `loss` not `signal_loss` and `signal_forward` or stuff like that. That doesn't make sense. Use the simplest name based on the context.

AI is very bad at creating this type of documentation page, we need more human intervention.

KomaMRICore/ext/KomaMRICoreReactantExt.jl
```diff
@@ -0,0 +1,22 @@
...
+end
+
+function _scalar_setindex!(x::TracedArray, value, i)
+    Reactant.allowscalar() do
+        x[i] = value
+    end
+    return x
+end
+
+end
```
cncastillo: I dont understand why we need any of this. Its probably not needed. Following my comment about redefining set/get.

KomaMRICore/src/simulation/SimMethods/Bloch/cpu/BlochCPU.jl
```diff
@@ -15,6 +15,9 @@ struct BlochCPUPrealloc{
     ΔBz::ΔBzType
 end
 
+@inline _scalar_getindex(x, i) = x[i]
+@inline _scalar_setindex!(x, value, i) = setindex!(x, value, i)
```
cncastillo: Is very unlikely we need this, just use regular indexing set and get.

KomaMRICore/src/simulation/SimMethods/Bloch/cpu/BlochCPU.jl
```diff
@@ -146,13 +149,13 @@ function run_spin_excitation!(
...
         #Effective field
-        B1 = seq.B1[i]
+        B1 = _scalar_getindex(seq.B1, i)
         @. Bz = (seq.Gx[i] * x + seq.Gy[i] * y + seq.Gz[i] * z) + ΔBz - seq.Δf[i] / T(γ) # ΔB_0 = (B_0 - ω_rf/γ), Need to add a component here to model scanner's dB0(x,y,z)
         @. B = sqrt(abs2(B1) + Bz^2)
         #Spinor Rotation
         @. φ_half = T(-π * γ) * (B * seq.Δt[i]) # TODO: Use trapezoidal integration here (?),  this is just Forward Euler
         @. α = cos(φ_half)
-        @. B = sin(φ_half) / (B + (B == 0) * eps(T))
+        @. B = (sin(φ_half) + (B == 0) * T(-π * γ) * seq.Δt[i]) / (B + (B == 0))
```
cncastillo: This change is incorrect, expressions are not identical.

KomaMRICore/src/simulation/SimulatorCore.jl
```diff
@@ -388,7 +388,8 @@ function simulate(
     if sim_params["gpu"]
         sim_params["Nthreads"] = 1
     end
-    sig = zeros(ComplexF64, Ndims..., sim_params["Nthreads"])
+    sig = similar(obj.ρ, ComplexF64, Ndims..., sim_params["Nthreads"])
+    fill!(sig, 0)
```
cncastillo, line 407: There is no need to change this. Also using `similar(obj.ρ, ComplexF64, ...` is sketchy, I thought `sig` was already a gpu array, why would it depend on `obj.ρ`. `|> gpu` is not used?

KomaMRICore/test/runtests.jl
cncastillo, line 1: What are you testing here? I really dislike this part

```julia
...
                probe = Base.invokelatest(
                    getfield,
                    @__MODULE__,
                    :run_reactant_enzyme_node_probe,
                )
                reactant_loss, reactant_gradient = Base.invokelatest(probe, params)
...
```
Why is `Base.invokelatest(` needed, this is VERY hacky and probably incorrect. If there is no justification for the test, I prefer not to test.

Let's add only tests that make sense, like showing we can optimize an RF using simulate and that the simulated profile matches, or something like that. Let's test functionality that we want preserved in time, not just the operations not failing.

cncastillo (changes requested):

KomaMRICore/src/simulation/SimulatorCore.jl
KookiesNKareem: The problem is that sig can't be a generic array if we're using reactant to trace something we eventually write to it, so in this case im trying to use obj.ρ as a way to get the appropriate type

cncastillo (changes requested):
Please dont add things that depend on `within_compile()`.

PR #931, opened 2026-08-18
Add documentation handling for old URLs referenced in paper

pvillacorta:
Preserve documentation URLs published in the [KomaMRI motion paper](https://doi.org/10.1002/mrm.70145).

## Changes

- Generate three legacy `/stable/` redirects from `docs/paper-links.toml` in the DocumenterVitepress build output, replacing handwritten redirect HTML and its copy loop.
- Keep the paper DOI inventory in TOML. GitHub-transfer and root-level Pages mappings remain commented out because this build does not manage them.
- Use maintained `/stable/` destinations, not a retained v0.14 snapshot. No per-release edits: if a paper-link destination moves, update its TOML mapping and verify the generated redirect, as required by `AGENTS.md`.

## Validation

- Full local docs build and targeted redirect checks passed.
- CI and the docs preview passed for `86d4400a6`; the new `a492e2956` commit only adds the paper-link-specific `AGENTS.md` rule.

The production `/stable/` aliases will be published with a release documentation deployment that includes this change.


pvillacorta:
Since https://github.com/JuliaHealth/KomaMRI.jl/commit/eacd462f0bcea277fa225cca2141cfb19157061d, only stable and dev versions are visible in the docs. Is this intended?

cncastillo:
> Since https://github.com/JuliaHealth/KomaMRI.jl/commit/eacd462f0bcea277fa225cca2141cfb19157061d, only stable and dev versions are visible in the docs. Is this intended?

yes, the repo was getting too big.

I think the approach of this pr is too hardcoded, and if a new stable version is release we will need to change all links. Can we make it so is slightly more robust?

pvillacorta:
My idea was to redirect the links to a fixed version (in this case v0.14), and leave it like that so we don't need to change them anymore every time a new version is released. That should work as long as that fixed version is preserved within the docs. I will investigate more, but I don't know whether I could make it more robust.

cncastillo:
If I need to update the link each time I release a new komamri (0.15 for example) I would not be able to merge this.

pvillacorta:
That is undoubtedly a better approach. What concerns me is that the redirect is set to /stable, rather than to a specific version of Koma. I say this because, for example, if another explanatory section were added, the GPU section might be renamed to something like 10-gpu-explanation, and that would have to be changed manually. By redirecting it to a fixed version of Koma instead (i.e., /v0.14/explanation), we wouldn’t run into that problem. The issue in this case would be that we’d have to maintain the build for that version on the gh-pages branch.

cncastillo:
I dont understand the problem, you just need to update the `paper-links.toml` if something changes.

Issue #625, opened 2025-10-08
Outflow Spin Reset at ADC points


mcgrathcm:
### What happened?

When discussing #620 , @cncastillo and I came across a possible issue with how `outflow_spin_reset` is called during ADC samples.

Currently, in `BlochCPU.jl , run_spin_precession!`,  `outflow_spin_reset` is called once per ADC point, and once at the end of the call. From my understanding, the one at the end of the call looks at all time points (`seq.t`) and resets _all_ spins that have seen a reset signal during this time. To me, this makes sense.

However, for ADC sample events, it appears to only check the current time point, which may or may not be a reset flag. The problem arises if a reset flag comes many samples before the ADC, but still within the same call of `run_spin_precession!`. Since the reset is only being checked at ADC calls and at the end of the function, this reset flag will be missed.

One solution could be to check the reset flag at every time step. Alternatively, the call of `outflow_spin_reset` during an ADC sample could check _all previous time points_, and if any contain a flag, reset that spin.

On a side note, it seems`BlochGPU.jl` is not performing any resets during ADC sampling?

Overall, while this does introduce unrealistic signal accumulation, I think the effect is rather minor, since it only should influence the signals near the inlet regions, which is often already heavily corrupted by transients due to the inflow of fresh magnetization.

### Environment

```Markdown
Discussion only, not tested
```

PR #935, opened 2026-08-21
Add periodic FlowPath magnetization remapping

### LINKED ISSUE #930: Add GPU magnetization remapping at periodic motion boundaries
## Problem

When a sequence exceeds the duration of a periodic motion definition, particle coordinates loop back to their initial positions. This works for closed motion, but not for flow trajectories: particles retain magnetization from unrelated end-of-cycle locations, corrupting spatially dependent steady state.

## Proposal

Precompute a nearest-neighbor map from each initial particle position to the closest end-of-cycle particle:

```julia
map[i] = argmin_j ‖r_end[j] - r_initial[i]‖
```

At each cycle boundary, remap both magnetization components:

```julia
M_next.xy[i] = M_end.xy[map[i]]
M_next.z[i]  = M_end.z[map[i]]
```

Multiple destinations may share a source, so this is a gather rather than necessarily a permutation.

## Implementation

- GPU only
- Insert a simulation time point at each cycle boundary and split the simulation block there.
- Before executing the next block, launch a standalone GPU gather kernel.
- Gather into a preallocated scratch `Mag` buffer, then swap buffers.
- Reuse the map and scratch buffer for every cycle.
- Keep the existing `FlowPath` equilibrium-reset behavior unchanged.

The remap should occur between the end-position and start-position sides of the motion discontinuity.

mcgrathcm:
## Summary

Adds GPU magnetization remapping for periodic, non-closed `FlowPath` motion.

- Accepts a precomputed destination-to-source `cycle_map`
- Splits simulation blocks at exact cycle boundaries
- Applies the map with a standalone GPU gather kernel and reusable scratch buffer
- Remaps both transverse and longitudinal magnetization
- Preserves existing inflow reset behavior
- Supports phantom serialization with and without a map
- Disallows phantom slicing when a map is present
- Reports a clear error for CPU simulation

Koma does not calculate the mapping; it must be supplied with the phantom.

## Testing

- `KomaMRIBase`: 775 tests passed
- `KomaMRICore` CPU: 437 tests passed
- `KomaMRIFiles`: 24,450 tests passed
- GPU behavior validated with a 2-million-spin periodic-flow simulation

Closes #930

### DIFF
--- a/KomaMRICore/src/simulation/Flow.jl
+++ b/KomaMRICore/src/simulation/Flow.jl
@@ -1,6 +1,66 @@
 spin_coordinates(motion, x, y, z, t) = get_spin_coords(motion, x, y, z, t)
 spin_coordinates(::NoMotion, x, y, z, t) = x, y, z
 
+struct CycleRemapWorkspace{I,M}
+    source::I
+    scratch::M
+end
+
+cycle_remap_indices(::Colon, N) = Base.OneTo(N)
+cycle_remap_indices(indices, N) = indices
+
+function cycle_remap_sources(obj::Phantom, motion::Motion)
+    affected = collect(cycle_remap_indices(KomaMRIBase.get_indexing_range(motion.spins), length(obj)))
+    map = motion.action.cycle_map
+    length(affected) == length(map) ||
+        throw(DimensionMismatch("cycle_map length must equal the number of affected phantom particles."))
+    source = collect(eachindex(obj.ρ))
+    source[affected] .= affected[map]
+    return source
+end
+
+function cycle_remap_break_indices(seqd, motion)
+    breaks = Int[]
+    for t in KomaMRIBase.cycle_remap_times(motion, last(seqd.t))
+        t < last(seqd.t) || continue
+        i = searchsortedlast(seqd.t, t)
+        i >= firstindex(seqd.t) && seqd.t[i] == t ||
+            error("Cycle boundary $t is missing from the simulation sampling grid.")
+        push!(breaks, i)
+    end
+    return breaks
+end
+
+@kernel function cycle_remap_kernel!(M_xy_new, M_z_new, @Const(M_xy), @Const(M_z), @Const(source))
+    i = @index(Global)
+    @inbounds begin
+        M_xy_new[i] = M_xy[source[i]]
+        M_z_new[i] = M_z[source[i]]
+    end
+end
+
+function cycle_remap_workspace(source, M::Mag, backend::KA.GPU)
+    return CycleRemapWorkspace(
+        gpu(source, backend),
+        Mag(similar(M.xy), similar(M.z)),
+    )
+end
+
+function remap_magnetization!(M::Mag, workspace::CycleRemapWorkspace, backend::KA.GPU, groupsize)
+    cycle_remap_kernel!(backend, groupsize)(
+        workspace.scratch.xy,
+        workspace.scratch.z,
+        M.xy,
+        M.z,
+        workspace.source;
+        ndrange=length(workspace.source),
+    )
+    KA.synchronize(backend)
+    M.xy, workspace.scratch.xy = workspace.scratch.xy, M.xy
+    M.z, workspace.scratch.z = workspace.scratch.z, M.z
+    return nothing
+end
+
 outflow_spin_reset_at!(spin_state, t, i, motion; replace_by=0) =
    outflow_spin_reset!(spin_state, t[i, :], motion; replace_by)
 outflow_spin_reset_at!(spin_state, t, i, ::NoMotion; replace_by=0) = nothing
--- a/KomaMRICore/src/simulation/Flow.jl
+++ b/KomaMRICore/src/simulation/Flow.jl
@@ -1,63 +1,33 @@
 spin_coordinates(motion, x, y, z, t) = get_spin_coords(motion, x, y, z, t)
 spin_coordinates(::NoMotion, x, y, z, t) = x, y, z
 
-struct CycleRemapWorkspace{I,M}
-    source::I
-    scratch::M
-end
-
-cycle_remap_indices(::Colon, N) = Base.OneTo(N)
-cycle_remap_indices(indices, N) = indices
-
-function cycle_remap_sources(obj::Phantom, motion::Motion)
-    affected = collect(cycle_remap_indices(KomaMRIBase.get_indexing_range(motion.spins), length(obj)))
-    map = motion.action.cycle_map
-    length(affected) == length(map) ||
-        throw(DimensionMismatch("cycle_map length must equal the number of affected phantom particles."))
+# Global particle index each particle takes its magnetization from at a cycle boundary.
+function cycle_remap_sources(obj, motion)
+    affected = KomaMRIBase.get_indexing_range(KomaMRIBase.expand(motion.spins, length(obj)))
     source = collect(eachindex(obj.ρ))
-    source[affected] .= affected[map]
+    source[affected] .= affected[motion.action.cycle_map]
     return source
 end
 
+# Cycle boundaries are motion key times, so discretize always samples them. The grid
+# value can differ from a freshly computed boundary by a few ulp, since it round-trips
+# through a per-block time offset.
 function cycle_remap_break_indices(seqd, motion)
     breaks = Int[]
     for t in KomaMRIBase.cycle_remap_times(motion, last(seqd.t))
         t < last(seqd.t) || continue
-        i = searchsortedlast(seqd.t, t)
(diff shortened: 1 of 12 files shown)


mcgrathcm:
Here is an example of the issue being solved for a spoiled GRE sequence. The artifact presents itself as a flash at the start the of the cycle, most noticeable at the outlet region. NN remapping performs nearest neighbor remapping of magnetization. The magnitude of this issue is very sequence dependent, so it may be more or less present depending on sequence. @cncastillo 
identity-vs-remapped-titled

mcgrathcm:
I have come across an issue that needs some investigation here.

Currently this implementation requires block splitting at periodic boundaries. This is fine for precession blocks, however for excitation blocks this requires **splitting the RF pulse**. I am not entirely sure how this will affect the simulation accuracy, especially for more advanced methods like BlochMagnus. 

Unfortunately I believe this is required, as this requires **communication of magnetization between spins**, which conflicts directly with the kernel architecture which is **one spin per thread**.

One upside, even if this somewhat breaks the calculation of the RF pulse, it likely will not break it as much as the current implementation where particles can teleport across the entire phantom, resulting in crazy state mixing.

Basically, the end result would mean that sometimes **a single RF pulse may split into two excitation blocks with non-zero start/end strength.** And in-between them there is a re-mixing of magnetization. 

@cncastillo @pvillacorta curious what you think.

pvillacorta:
Hi! Sorry for the delay in replying. I think this RF-splitting is worthwhile, as long as this behavior in the simulation is isolated to the FlowPath case.

I still need to fully understand this specific issue and see how it fits into some of the other issues we have already addressed. If I understand correctly, in #620 we identified two related problems involving periodic motion and flow:

1. **(Blue dot in the diagram in [#620](https://github.com/JuliaHealth/KomaMRI.jl/issues/620))** Both when spin remapping occurred and when a new motion period started, the magnetization of the affected spins was reset to their initial magnetization (M0). However, there were cases where, when discretizing the sequence, no key point was created within the reset window (while the `spin_reset` flag was set to 1). This is why, in #638, we forced additional points to be added to the `DiscreteSequence` near these boundaries.

2. **(#930, yellow dot in the diagram in [#620](https://github.com/JuliaHealth/KomaMRI.jl/issues/620))** There are cases where the spin magnetization after being "reset" should not be equal to the initial M0, but should instead start from an intermediate state between 0 and M0 (e.g. M_steady_state). To solve this, we would need to be able to set the spin magnetization, after the reset, to an arbitrary value. We had not solved this problem until now.
I have a couple of clarifying questions:
   1. What you are proposing is to set this arbitrary value to that of the nearest neighbor when each motion period ends and a new one begins, correct?
   2. Is this only needed when a new period starts, or is it also needed when a spin leaves an outlet and enters an inlet (`spin_reset=1`)? I find it difficult to distinguish between these two situations. From what I can see in the GIFs, NN remapping is only needed in the first case, right? If there is a reset within a period, could we assume that the initial state of that spin should go back to M0, since it is effectively "coming from outside" the excited slice?
  
3. There is also this other related issue, #625, which is still unresolved.

I want to reiterate that I think this solution is worthwhile because it simulates the behavior much more accurately.

Just to clarify one more point:

> One upside, even if this somewhat breaks the calculation of the RF pulse, it likely will not break it as much as the current implementation where particles can teleport across the entire phantom, resulting in crazy state mixing.

The particles are still going to teleport, right? It's just that their magnetization will be able to acquire more accurate values when the period is reset. So, just to make sure I understand the proposed solution correctly.

Thanks!

mcgrathcm:
Yes, this directly addresses #620. 

1. For adding time points at spin reset events, #638 seems to be working well and I do not see any issues. Interestingly for the large scale cases we are using, this ends up adding a simulation time point at every motion point, as there is constant resetting.
2. It is basically what you have stated in ii, this PR is specifically addressing the periodic wrapping of the entire `FlowPath`, not the intermediate usage of `spin_reset`. What you are pointing to here is a real issue, but solving this is non-trivial, as we would need to determine the correct steady state signal (including phase if we want to set Mxy). I think this is out of scope for the simulation package. Maybe in the future we can allow the user to set the "steady state" M, but the responsibility to determine the correct value would be on them.
3. I have somewhat forgotten about this, but will check it out again and see if I can incorporate a fix here.

And to your final comment, yes the spins will still teleport, and thanks to your changes in #638 the teleport is over `MIN_RISE_TIME` so negligible phase will be added.
The flow is:
- Particles reach final step in `FlowPath`
- Particles take a single step of `MIN_RISE_TIME` back to starting point, negligible change to magnetization
- Particle magnetization is re-mapped such that the ensemble magnetization structure of the flow is preserved.
- Continue normally

There are a few minor fixes I still need to add, but otherwise this seems to be functional on my end.

mcgrathcm:
At the current time, this PR is functional and should be fully backwards compatible. If the user does not provide a remapping, it should behave as before. If this is provided, the remapping is performed at each periodic cycle. 
Supports both CPU and GPU.

cncastillo:
>If the user does not provide a remapping, it should behave as before

did you try loading a "new" `.phantom` with `Flow` motion in an old version of koma? "old" `.phantom` with new `Flow`? Is KomaMRIFiles even able to write the spin remapping?

We might need to set this as a breaking change depending on the results of the read/write tests.

mcgrathcm:
I have tested all permutations of reading and writing between this PR and master, and it all works. 
One thing to note, when the older version of Koma reads a phantom with this new remap, it will silently drop the remap. 
We could make this a major version bump, in which case the reader would produce a warning when loading a new phantom.

cncastillo:
> Koma reads a phantom with this new remap, it will silently drop the remap.

thats perfect, then a patch bump is reasonable as it just adds adds stuff and it doesn't break things.

@pvillacorta can you review this PR?

KomaMRICore/src/simulation/SimulatorCore.jl
```diff
@@ -357,12 +360,17 @@ function simulate(
     end
     # Simulation init
     seqd = discretize(seq; sampling_rule, motion=obj.motion, freq_in_phase=sim_params["freq_in_phase"]) # Sampling of Sequence waveforms
+    remap_motion = KomaMRIBase.cycle_remap(obj.motion)
+    remap_breaks = isnothing(remap_motion) ? Int[] : cycle_remap_break_indices(seqd, obj.motion)
+    remap_sources = isnothing(remap_motion) ? nothing : cycle_remap_sources(obj, remap_motion)
```
pvillacorta: I think that this could be done better with method dispatch. I will try to do it this afternoon.

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -1,6 +1,36 @@
 spin_coordinates(motion, x, y, z, t) = get_spin_coords(motion, x, y, z, t)
 spin_coordinates(::NoMotion, x, y, z, t) = x, y, z
 
+# Global particle index each particle takes its magnetization from at a cycle boundary.
+function cycle_remap_sources(obj, motion)
+    affected = KomaMRIBase.get_indexing_range(KomaMRIBase.expand(motion.spins, length(obj)))
+    source = collect(eachindex(obj.ρ))
+    source[affected] .= affected[motion.action.cycle_map]
+    return source
+end
```
pvillacorta: An example of why I would use method dispatch (on action's type).
For this case, since not all actions have a `cycle_map` field, the function will break if a motion containing an action different from `FlowPath` is passed as input.

pvillacorta (commented):
Since I did not want to directly touch this PR, I created [this one](https://github.com/mcgrathcm/KomaMRI.jl/pull/4) (after a few failed attemps :sweat:) in order to apply some minor changes regarding method dispatch and so on. 
Maybe @mcgrathcm can you test it with those big phantoms?

KomaMRICore/src/simulation/Flow.jl
mcgrathcm: In its current form, this function is only reachable if `cycle_map` is present. Not sure what the design philosophy is here, if we want to protect against future code changes or if we leave it minimal and future code changes must implement dispatch as needed. @cncastillo what is correct?

KomaMRICore/src/simulation/SimulatorCore.jl
mcgrathcm: I agree it could be better with method dispatch, but keep in mind this needs to be backwards compatible, so if an old phantom is provided that uses `FlowPath` but does not contain `cycle_map`, this should default to the older behavior and not do any of this remapping stuff.

mcgrathcm:
Thanks @pvillacorta , I have merged your changes, and tested that they produces identical results. 
At this point, the feature is working as expected, the only thing left is and CI/CD changes needed.

PR #946, opened 2026-09-09
Docs/Fix scanner examples for the HardwareLimits API

JanWP:
There are a few places in the sequence creation and Pulseq translation guides that have not been migrated to the recently introduced HardwareLimits API. Use Scanner(limits=HardwareLimits(...)) and sys.limits.GR_Δt, including the Unitful example.

This is a rather trivial change. The issue was spotted and the fix proposed by Codex. I checked that the modified code
snippets run fine locally in the test environment. I did not run a documentation build.


cncastillo:
Oh true! Thanks!

JanWP:
I accidentally included an unrelated change in this PR. This was supposed to only contain the documentation commit.

PR #947, opened 2026-09-09
Docs/Fix scanner examples for the HardwareLimits API - v2

JanWP:
After my previous botched PR, this should be better.

There are a few places in the sequence creation and Pulseq translation guides that have not been migrated to the recently introduced HardwareLimits API. Use Scanner(limits=HardwareLimits(...)) and sys.limits.GR_Δt, including the Unitful example.

This is a rather trivial change. The issue was spotted and the fix proposed by Codex. I checked that the modified code
snippets run fine locally in the test environment. I did not run a documentation build.

### DIFF
--- a/docs/src/how-to/3-create-your-own-sequence.md
+++ b/docs/src/how-to/3-create-your-own-sequence.md
@@ -186,7 +186,7 @@ limits.
 == KomaMRI
 
 ```julia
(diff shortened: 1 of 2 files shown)


cncastillo:
No worries! Thanks! I will merge as soon as tests pass.

PR #948, opened 2026-09-10
Floating point failure for periodic motion

mcgrathcm:
## What

`add_period_times!` brackets every period boundary with a sample at `t ± MIN_RISE_TIME`
(1e-14 s). For a periodic motion this is what confines the position discontinuity at the
wrap to a ~10 fs interval, so essentially no phase accrues across it no matter how far a
spin jumps.

Those offsets are applied to the **base** period times. `extend_periodic!` then shifts them
out across the sequence by adding `k * period`. Above roughly 90 s, 1e-14 is smaller than
half a Float64 ulp, so the offset rounds away during that shift and both bracketing samples
collapse onto the boundary itself.

`motion_sampling_times` then dedups them, the boundary becomes a single grid point, and the
position jump gets spread over a whole simulation step instead of 10 fs — producing a large
spurious phase, silently and with nothing thrown.

## How

Compute the offset once in `add_key_time_points!`, sized at the largest time the run
actually reaches rather than at the base period times, and pass it to `add_period_times!`:

```julia
rise_time = max(MIN_RISE_TIME, 8eps(float(t_max)))

### DIFF
--- a/KomaMRIBase/src/motion/Motion.jl
+++ b/KomaMRIBase/src/motion/Motion.jl
@@ -255,7 +255,8 @@ function add_key_time_points!(t, a, t_start::T, t_end::T, periods, periodic) whe
     aux = T[] 
     period = sum((t_end - t_start) .* periods)
     t_max = maximum(t)
-    add_period_times!(aux, t_start, t_end, periods)
+    rise_time = max(MIN_RISE_TIME, 8eps(float(t_max)))
+    add_period_times!(aux, t_start, t_end, periods, rise_time)
     add_reset_times!(aux, a, t_start, t_end, periods)
     extend_periodic!(aux, t_max, period, Val(periodic))
     append!(t, aux[aux .<= t_max])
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: KomaMRIBase/src/motion/Motion.jl [hunk 1]
```diff
@@ -255,7 +255,8 @@ function add_key_time_points!(t, a, t_start::T, t_end::T, periods, periodic) whe
     aux = T[] 
     period = sum((t_end - t_start) .* periods)
     t_max = maximum(t)
-    add_period_times!(aux, t_start, t_end, periods)
+    rise_time = max(MIN_RISE_TIME, 8eps(float(t_max)))
```
cncastillo: `8eps(float(t_max))` let's avoid magic numbers like 8.

cncastillo (changes requested):

KomaMRIBase/src/motion/Motion.jl
mcgrathcm: I have shrunk this to `4eps`, but that is the minimum recommended as another change in #935 uses a `+-2eps` tolerance for finding when to perform the remapping, and we need to avoid it catching the incorrect point. 

Technically using `2eps` here and `+-1eps` in #935 would work but provides exactly zero margin for additional errors

cncastillo: Can we at least give it a name to the constant so its clear what its representing/mean?

cncastillo: Why do you need `float(t_max)`?

mcgrathcm:
I have reverted previous changes in favor of only updating `MIN_RISE_TIME` from `1e-14` to `1e-10`. 
`1e-10` is the largest round value that leaves every other consumer untouched:
- `HardwareChecks` uses `max(100 * MIN_RISE_TIME, sqrt(eps))`, which stays dominated by `sqrt(eps)` up to `1.49e-10`
- `MAX_STEP_TIME_SNAP_TOL` becomes `1e-11`, still five orders below the 1 µs Pulseq RF raster.
- The wider bracket accrues ~5e-5 rad across a 10 cm jump in a 40 mT/m gradient.

cncastillo:
> MAX_STEP_TIME_SNAP_TOL becomes 1e-11, still five orders below the 1 µs Pulseq RF raster.

This change affects Pulseq read/write, ideally the changes would only affect the motion part.

PR #957, opened 2026-09-18
Outflow spin reset fix

### LINKED ISSUE #625: Outflow Spin Reset at ADC points

mcgrathcm:
## Summary

Fixes #625 by tracking spin resets cumulatively within each simulation block. Spins are now reset if any reset flag occurs before the current timestep or ADC sample, including during RF excitation.

This adds `SpinResetState`, a preallocated per-block state containing:

- `first_reset`: the earliest reset timestep for each spin.
- `has_reset`: the running reset state used by CPU methods.
- `reset_column_ids` and `sampled_reset_columns`: reusable buffers for mapping simulation timesteps to `FlowPath.spin_reset` columns.

The GPU kernels receive `first_reset` directly and perform the reset within their timestep loops. This avoids constructing a dense spin-by-time mask and ensures that reset spins remain at equilibrium for the rest of the block.

The new behavior is used by Bloch, BlochSimple, and BlochMagnus CPU and GPU simulations.

@pvillacorta and @cncastillo please review.


pvillacorta:
Thank you! I will review this and #955 on Monday

KomaMRICore/src/simulation/Flow.jl
```diff
@@ -1,6 +1,172 @@
...
+   spin_state, state::SpinResetState, step, t, motion; replace_by=0,
+)
+   advance_spin_reset!(state, step)
+   outflow_spin_reset!(spin_state, state; replace_by)
+   return nothing
+end
+function outflow_spin_reset_at!(
+   spin_state, ::NoSpinReset, step, t, motion; replace_by=0,
+)
+   outflow_spin_reset_at!(spin_state, t, step, motion; replace_by)
```
pvillacorta: Why do we need to add this line if state is `NoSpinReset`?

pvillacorta (commented):
All changes make sense to me. However, don't you think we could unify `outflow_spin_reset!` methods so that all simulation methods use the new ones? This applies to BlochSimple and BlochDict, with the last of them kept unchanged for now.

KomaMRICore/src/simulation/Flow.jl
mcgrathcm: For unifying, I believe `BlochSimple` uses a slightly different internal representation, so it requires a little wrapper for calling `outflow_spin_reset!`, but it already shares the underlying functions. 
For `BlochDict` I talked to @cncastillo and we agreed that maybe we stop supporting it. For this PR supporting `BlochDict` would be more trouble than it's worth.

Good catch on the other comment, I have simplified it.

